PR #55, opened 2015-08-26
Add support for systemd service Type=notify.

mlang:


DaveMielke:
Tracked in the systemd branch until configure issues are resolved.

PR #60, opened 2015-11-03
memcpy and mempcpy expect a nonnull argument. (ml)

mlang:
This fixes a few warnings emitted by the LLVM/Clang static analyzer.


DaveMielke:
I'm not sure about this one. In the braille driver cases where the mempcpy() 
source pointer is NULL, the associated byte count is 0. The tests in the 
braille drivers, therefore, seem entirely useless and cluttering. The mempcpy() 
in usb_linux is fine except that the analyzer isn't figuring that out. Having a 
NULL test for that one actually makes the code confusing.

PR #73, opened 2016-09-05
Initial support for Handy Tech Actilino. (ml)

mlang:


DaveMielke:
The Actilino model entry still specifies the as40 key table.

DaveMielke:
I've made the change in master.

PR #76, opened 2016-10-22
AppStream requires uppercase hexadecimal vendor and product IDs. (ml)

mlang:


DaveMielke:
Looks like I did the very same thing just before receiving your pull request.

PR #83, opened 2016-12-29
brltty Android support request

discapacidad5:
Excuse my bad english use google translator

I was looking at brailleback that is for android and it is developed under brltty and works wonders but only with lines by bluetooth

Could you create a brailleback modification so that it recognizes braille lines per usb and serial port?

Although it managed to connect brltty with the android does not work well since the focus does not follow the braille line and for that reason in android you can not look down in a list used navigation braille brailleback line this works wonders

But with brailleback I can not take advantage of my old braille line that does not have bluetooth

I think this will help thousands of people who have old braille lines for how expensive it is to get a new one

Thanks for reading me sorry if the publication does not use it correctly but I do not sew how to write here


DaveMielke:
[quoted lines by discapacidad5 on 2016/12/28 at 17:54 -0800]

I'll answer your questions here, but it'd be far better for you to subscribe to 
brltty's mailing list. Go to brltty.com, click on Contact, and follow the 
instructions.

>Could you create a brailleback modification so that it recognizes braille 
>lines per usb and serial port?

No, because we don't maintain BrailleBack. I'm not even sure that Google still 
maintains it. We maintain brltty, and it does recognize USB devices on Android.

>Although it managed to connect brltty with the android does not work well 
>since the focus does not follow the braille line and for that reason in 
>android you can not look down in a list used navigation braille brailleback 
>line this works wonders

Yes, it does, but you just need to know how to do it. Brltty places the cursor 
where the focus is. You can then move around the screen without changing the 
focus, but, by pressing the leftmost routing key you can bring the focus to 
wherever you are.

PR #84, opened 2017-01-03
Build fixes for Solaris

alanc:
These are patches we've been using to build the brltty-5.4 packages for Solaris.

### DIFF
--- a/Drivers/Braille/EuroBraille/eutp_brl.c
+++ b/Drivers/Braille/EuroBraille/eutp_brl.c
@@ -5,7 +5,9 @@
 ** Login   <obert01@epita.fr>
 */
 
+#ifndef _XOPEN_SOURCE
 #define _XOPEN_SOURCE 500
+#endif
 
 /* globals */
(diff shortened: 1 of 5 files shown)


DaveMielke:
Thanks. In configure.ac, I changed solaris2.1? to solaris2.1?* - I hope that's 
okay.

alanc:
That's fine - I'd used ? just to skip over Solaris 2.1, but since that far predates the required XPG5 support it's not going to matter anyway.  I pulled your changes and it still builds okay for me here.

DaveMielke:
Does Solaris have the ability to directly inspect the screen content so that we 
can look into writing a native screen driver for it? I recall that there was 
talk of doing that several years ago. Also, virtual consoles.

alanc:
By "screen content" I assume you mean in tty mode?  (Since graphics mode is just the X Window System like most other Unixes.)  I don't know and would have to investigate it.

Virtual consoles are supported in Solaris 11 and later - the ioctls for them are documented on http://docs.oracle.com/cd/E53394_01/html/E54777/vt-7i.html

PR #86, opened 2017-01-31
Preliminary driver for the DOT pad prototype (http://dotincorp.com/)

minorninth:
Hi - this is a driver for a prototype board from DOT (http://dotincorp.com/). The prototype doesn't have any input yet, it just supports multi-line output.

I'm sure my change to usb_serial.c is not right. Can you help me figure out the right way to do that?


minorninth:
This is ancient, closing.

PR #87, opened 2017-01-31
Preliminary driver for Bristol Canute multi-line braille display.

minorninth:
Oh, I think I may have figured out how to identify it as a serial device. How's this?


sandeepmistry:
I suggest this patch not be applied, the Bristol Canute, is incorrectly using the same USB vendor and product ID as the [Arduino Leonardo](https://www.arduino.cc/en/Main/ArduinoBoardLeonardo).

Newer version of Chrome OS ship with this patch (https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/master/app-accessibility/brltty/files/brltty-5.4-canute.patch) and it prevents the Arduino Leonardo as appearing as a serial port, so it can't be used with Chrome apps.

I've also reported the issue with the Chromium team: https://bugs.chromium.org/p/chromium/issues/detail?id=739176

DaveMielke:
Is this a different version of the driver than what's already in brltty's 
bristol branch? If so, would you mind comparing them? I've committed your pull 
into brltty's canute branch.

sandeepmistry:
Latest news from the Chromium side is that the incorrect USB ID's were used: https://bugs.chromium.org/p/chromium/issues/detail?id=739176#c2

PR #93, opened 2017-04-06
Add support for "EsysLight80" and "EsytimeEvolution"

dpoitou:
EsysLight80 is a new braille display likes Esys 80 but without keyboard.
Esytime Evolution is a new braille note likes Esytime but with bluetooth adapter.

These modifications allow brltty to work fine with these devices via USB port.

Thanks a lot to add this contribution.


DaveMielke:
Thanks. Just in time for our 5.5 release.

PR #109, opened 2017-12-10
Implement brlapi_getDeviceIdentifier().

shindere:


DaveMielke:
What does it return? I'm guessing that it should perhaps be called 
getModelIdentifier(). A device identifier sounds like something that'd be 
unique to a specific device.

-- 
Dave Mielke           | 2213 Fox Crescent | WebHome: http://Mielke.cc/
EMail: Dave@Mielke.cc | Ottawa, Ontario   | Twitter: @Dave_Mielke
Phone: 1-613-726-0014 | Canada  K2A 1H7   |

shindere:
Dave Mielke (2017/12/10 21:57 +0000):
> What does it return?

The content of the keyBindings field of the braille display sutructure.

> I'm guessing that it should perhaps be called 
> getModelIdentifier(). A device identifier sounds like something that'd be 
> unique to a specific device.

Yeah that's how it's actually called: the error is in the commit
message, sorry about that.

PR #113, opened 2017-12-29
Remove MKOBJ in favour of MKMOD. (ml)

mlang:
Calling ld directly can lead to problems when cross-compiling.
MKMOD *should* work on other platforms as well, as we are already using it to
link braille.o for all other drivers.  I couldn't figure out what the
motivation for a separate helper was, please let me know if
I have missed something.


DaveMielke:
The change is fine. Thanks. MKOBJ was for creating incomplete object files, and 
MKMOD was for creating shared object files. Of course, they both end up being 
the smae sort of operation. MKOBJ would've only ever been used in two places - 
the EuroBraille driver and the Vario driver. The latter disappeared a long time 
ago when the Baum driver was created. That left the single case of the 
EuroBraille driver, which you finally discovered.

DaveMielke:
Thinking about it some more: MKOBJ was first and used in both cases. Then, Mac 
OS X made it necessary to create shared objects in a better way so MKMOD (Mac 
OS X refers to loadable shared objects as modules) came into existence. That 
left MKOBJ still being used for the incomplete object case, which wasn't 
cleaned up till today.

PR #120, opened 2018-02-17
Advise the programmer to use brlapi_setExceptionHandler if she'd like to

sthibaul:


shindere:
Minor, but woldn't "Aborting" be more precise and carrying more
information than "Crashing"?

sthibaul:
I don't mind either way :)

shindere:
Samuel Thibault (2018/02/17 15:24 +0000):
> I don't mind either way :)

Then I'm for abort.
Thanks.

>

PR #149, opened 2018-10-03
Fix linking error on mips64el

ffontaine:
Patch that was fixing build on mips64el was reverted:
https://github.com/brltty/brltty/commit/9e7d62c869d3c1cbe12dda8b0291a4692c193416

So fix the issue again, this time by checking if $(GCC) is available as
suggested by Arnout during review of buildroot patch:
https://patchwork.ozlabs.org/patch/972614

Fixes:
 - http://autobuild.buildroot.org/results/31f682838b3d3b2c7103b5c51f2aba0b89d4f630

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -481,7 +481,12 @@ AC_CACHE_CHECK([for make relocatable object command], [brltty_cv_prog_make_objec
 case "${host_os}"
 in
    *)
-      brltty_cv_prog_make_object_relocatable="\$(LD) -r -o"
+      if test "${GCC}" = "yes"
+      then
+         brltty_cv_prog_make_object_relocatable="\$(CC) -shared -o"
+      else
+         brltty_cv_prog_make_object_relocatable="\$(LD) -r -o"
+      fi
       ;;
 esac])
 MKREL="${brltty_cv_prog_make_object_relocatable}"


sthibaul:
Is this still an issue? In Debian we have been building mips64el versions fine without changes.

ffontaine:
We still have this patch in buildroot: https://git.buildroot.net/buildroot/commit/package/brltty?id=9143c217213542574586c0e6a9f003e822633917.

It was added to fix the numerous build failures that we got with brltty 5.6 at this time: http://autobuild.buildroot.org/?reason=brltty-5.6:
```

/home/buildroot/buildroot-test/instance-1/output/host/bin/mips64el-ctng_n64-linux-gnu-ld -r -o braille.o eu_braille.o eu_clio.o eu_esysiris.o
/home/buildroot/buildroot-test/instance-1/output/host/bin/mips64el-ctng_n64-linux-gnu-ld: eu_braille.o: ABI is incompatible with that of the selected emulation
/home/buildroot/buildroot-test/instance-1/output/host/bin/mips64el-ctng_n64-linux-gnu-ld: failed to merge target specific data of file eu_braille.o
/home/buildroot/buildroot-test/instance-1/output/host/bin/mips64el-ctng_n64-linux-gnu-ld: eu_clio.o: ABI is incompatible with that of the selected emulation
/home/buildroot/buildroot-test/instance-1/output/host/bin/mips64el-ctng_n64-linux-gnu-ld: failed to merge target specific data of file eu_clio.o
/home/buildroot/buildroot-test/instance-1/output/host/bin/mips64el-ctng_n64-linux-gnu-ld: eu_esysiris.o: ABI is incompatible with that of the selected emulation
/home/buildroot/buildroot-test/instance-1/output/host/bin/mips64el-ctng_n64-linux-gnu-ld: failed to merge target specific data of file eu_esysiris.o
/home/buildroot/buildroot-test/instance-1/output/host/bin/mips64el-ctng_n64-linux-gnu-ld: Attempt to do relocatable link with elf64-tradlittlemips input and elf32-ntradlittlemips output
/home/buildroot/buildroot-test/instance-1/output/host/bin/mips64el-ctng_n64-linux-gnu-ld: eu_braille.o: file class ELFCLASS64 incompatible with ELFCLASS32
/home/buildroot/buildroot-test/instance-1/output/host/bin/mips64el-ctng_n64-linux-gnu-ld: final link failed: File in wrong format
make[3]: *** [braille.o] Error 1
make[3]: Leaving directory `/home/buildroot/buildroot-test/instance-1/output/build/brltty-5.6/Drivers/Braille/EuroBraille'
```

Since then, we bumped to 6.0. Following your feedback, I tried to remove this patch. Unfortunately, I can confirm that this patch is still needed.

sthibaul:
I guess the difference with Debian is the `ctng_n64`part?

The `ABI is incompatible with that of the selected emulation` error message looks to me like an incoherency between what gcc produces and what ld expects, i.e. rather a bug in the toolchain.

Using `gcc -shared` instead of `ld -r` is questionable, because the wording is not really the same in the respective documentations:

- `gcc -shared`: Produce a shared object which can then be linked with other objects to form an executable.
- `ld -r`: Generate relocatable output---i.e., generate an output file that can in turn serve as input to ld.

and the results indeed looks very different e.g. on x86:

```
file test1.o test2.o
gcc-shared-output.o: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, BuildID[sha1]=c52c1208cfeb208b1b8704588d04ed619355a37e, not stripped
ld-r-output.o: ELF 64-bit LSB relocatable, x86-64, version 1 (SYSV), not stripped
```

The first one is already linked with the libc library, has phdr, GOT, etc.

ffontaine:
These build failures were (and still are) raised with slightly different toolchains:
- http://autobuild.buildroot.org/toolchains/tarballs/mips64el-ctng_n64-linux-gnu.tar.xz
- http://autobuild.buildroot.org/toolchains/tarballs/br-mips64r6-n64-el-hf-2018.05.tar.bz2
- http://autobuild.buildroot.org/toolchains/tarballs/br-mips64-n64-full-2018.05.tar.bz2

We were also a bit confused about this issue during this patch review, you can see our full discussion here: https://patchwork.ozlabs.org/patch/972614

At that time, I tried to understand why on these toolchains, by default, ld wanted to link a 32 bits binary.  I found that an other possible worakaround would have been to set LD to `$(TARGET_LD) -melf64ltsmip`

Arnout explained me that
> in general, it is possible that additional options have to be passed to ld to
> make sure it creates the correct type of ELF file. The buildroot toolchain-wrapper makes
> sure that the correct options are passed but as brltty calls ld directly so it bypasses those.

If you think that this PR is not right, I can try to replace it by the "LD" workaround. I think it could be ok as we generally don't like to keep non upstreamable solution.

sthibaul:
brltty doesn't call a mere `ld`, but `mips64el-ctng_n64-linux-gnu-ld`, on objects which were created with `mips64el-ctng_n64-linux-gnu-gcc`. So that should be "just working" if the toolchain was correctly set up? I mean, isn't the triplet prefix exactly supposed to properly generate coherent compilation and linking details?

DaveMielke:
[quoted lines by Fabrice Fontaine on 2020/04/03 at 14:56 -0700]

>If you think that this PR is not right, I can try to replace it by the "LD" workaround. I think it could be ok as we generally don't like to keep non upstreamable solution.

it most assuredly isn't right. Without even looking, I can guarantee to you
that it isn't even working correctly in your case. The absence of an error
message doesn't mean a good build. It's just that you'd only see the problem if
your users were using certain braille devices that use that particular macro.

autoconf should be setting $LD to the correct version of ld within the
toolchain. If it isn't then it might be that you're setting $LD to ld before
the configure. As long as $LD isn't preset then autoconf should be doing the
right thing.

-- 

DaveMielke:
Is it possible that it used to be a problem but has been fixed since back
whenever and you just hadn't noticed and continued to assume that the patch was
still necessary? It really should be working correctly now due to this line in
configure.ac:

   AC_PATH_TOOL([LD], [ld])

In earlier days, ld was searched for via AC_PATH_PROG() instead but that's been
corrected for a while.

PR #182, opened 2019-09-01
Add support for the middle dot character to the fr-vs text table

shindere:


shindere:
Dave Mielke (2019/09/01 07:35 -0700):
> Merged #182 into master.

Thanks!

PR #190, opened 2019-11-04
brlapi parameters: add type for each parameter

sthibaul:
and use them in server and apitest


DaveMielke:
Maybe you could rebase brlapi-parameters.

PR #234, opened 2019-11-29
fixes from coverity

sthibaul:


sthibaul:
I have updated the MR with a few more fixes.

PR #264, opened 2020-04-25
libbrlapi: Automatically get VT nr from logind

sthibaul:
Notably on gnome with Wayland, XDG_VTNR may not be getting defined. We
can however just directly ask logind which VT our process is running on.

### DIFF
--- a/Programs/brlapi_client.c
+++ b/Programs/brlapi_client.c
@@ -92,6 +92,10 @@
 #define MAXIMUM_VIRTUAL_CONSOLE 1
 #endif /* MAXIMUM_VIRTUAL_CONSOLE */
 
+#ifdef HAVE_SD_SESSION_GET_VT
+#include <systemd/sd-login.h>
+#endif /* HAVE_SD_SESSION_GET_VT */
+
 #define BRLAPI_NO_DEPRECATED
 #include "brlapi.h"
 #include "brlapi_protocol.h"
@@ -1602,9 +1606,64 @@ int BRLAPI_STDCALL brlapi__enterTtyModeWithPath(brlapi_handle_t *handle, int *tt
 
   /* OK, Now we know where we are, so get the effective control of the terminal! */
   ttytreepath = getenv("WINDOWPATH");
-  if (!ttytreepath && getenv("DISPLAY"))
+  if (!ttytreepath && (getenv("DISPLAY") || getenv("WAYLAND_DISPLAY"))) {
     /* Cope with some DMs which are not setting WINDOWPATH (e.g. gdm 3.12) */
     ttytreepath = getenv("XDG_VTNR");
+
+#ifdef HAVE_SD_SESSION_GET_VT
+    if (!ttytreepath) {
+      /* Fallback to asking logind */
+      char *id;
+      unsigned vtnr = 0;
+      int ret;
+
+      ret = sd_pid_get_session(0, &id);
+      if (ret == 0 && id != NULL) {
+	sd_session_get_vt(id, &vtnr);
+	free(id);
+      } else {
+	char **sessions;
+	/* Not even logind knows :/ we are probably logged in from gdm */
+	ret = sd_uid_get_sessions(getuid(), 0, &sessions);
+
+	if (ret > 0) {
+	  int i, chosen = -1;
+
+	  for (i = 0; i < ret; i++) {
+	    char *type;
+
+	    ret = sd_session_get_type(sessions[i], &type);
+	    if (ret == 0) {
+	      if (strcmp(type, "tty") != 0 &&
+	          strcmp(type, "unspecified") != 0) {
+		/* x11, wayland or mir, so graphical.
+		 * User normally has only one of them */
+		if (chosen >= 0) {
+		  /* Oops, several sessions? That is not supposed to happen, better
+		   * choose none of them. */
+		  chosen = -1;
+		  break;
+		}
+		chosen = i;
+	      }
+	    }
+	  }
+
+	  if (chosen >= 0) sd_session_get_vt(sessions[i], &vtnr);
+
+	  for (i = 0; i < ret; i++) free(sessions[i]);
+	  free(sessions);
+	}
+      }
+
+      if (vtnr) {
+	size_t size = sizeof(vtnr)*3 + 1;
+	ttytreepath=alloca(size);
+	snprintf(ttytreepath, size, "%u", vtnr);
+      }
+    }
+#endif /* HAVE_SD_SESSION_GET_VT */
+  }
 
   nttytreepath = 0;
   if (ttytreepath) {
--- a/config.h.in
+++ b/config.h.in
@@ -254,6 +254,9 @@ extern "C" {
 /* Define this if the function shl_load is available. */
 #undef HAVE_SHL_LOAD
 
+/* Define this if the function sd_session_get_vt is available in libsystemd. */
+#undef HAVE_SD_SESSION_GET_VT
+
 /* Define this to be a string containing the copyright notice. */
 #undef PACKAGE_COPYRIGHT
 
--- a/configure.ac
+++ b/configure.ac
@@ -1230,6 +1230,17 @@ in
       ;;
 esac
 
+BRLTTY_HAVE_PACKAGE([systemd], [libsystemd], [
+    brltty_libs_save="${LIBS}"
+    LIBS="${LIBS} ${systemd_libs}"
+    AC_CHECK_FUNC([sd_session_get_vt], [
+      AC_DEFINE([HAVE_SD_SESSION_GET_VT], [1],
+                [Define this if the function sd_session_get_vt is available in libsystemd.])
+      api_libraries="${api_libraries} ${systemd_libs}"
+    ])
+    LIBS="${brltty_libs_save}"
+])
+
 BRLTTY_ARG_PACKAGE([params], [boot parameters], [], [dnl
    linux*)
       params_package="linux"


Programs/brlapi_client.c
```diff
@@ -1602,10 +1606,29 @@ int BRLAPI_STDCALL brlapi__enterTtyModeWithPath(brlapi_handle_t *handle, int *tt
...
     /* Cope with some DMs which are not setting WINDOWPATH (e.g. gdm 3.12) */
     ttytreepath = getenv("XDG_VTNR");
 
+#ifdef HAVE_SD_SESSION_GET_VT
+    if (!ttytreepath) {
+      /* Fallback to asking logind */
+      char *id;
+      unsigned vtnr;
+      int ret = sd_pid_get_session(0, &id);
+      if (ret == 0 && id != NULL) {
```
benzea, line 1619: Unfortunately, this will not be enough in all cases. If `sd_pid_get_session` returns an error, then we need to start guessing the session. Which requires looping over all sessions and finding the "best" graphical one.

I can cook up a patch if that is helpful.

sthibaul: I don't think we want to guess, that most often provides more harm than good :/

benzea: Yup, unfortunately we really have to. There is no way around it and it is safe because users can only have one graphical session by definition. Some people have not accepted that one yet though.

The thing is that the processes of the graphical session end up being started by systemd. And then they technically don't actually belong to any user session. There are some further long term ideas and more details … but right now, guessing is definitely the right thing to do!

benzea: Note that you *should* use `sd_pid_get_session` if it returns something sane. Only when that fails fall back to guessing.

sthibaul: > unfortunately we really have to

Why? In which situation sd_pid_get_session might not find it?

> users can only have one graphical session by definition

?

one can run startx several times on different VTs. We do not want to mix Braille output in that case.

> The thing is that the processes of the graphical session end up being started by systemd. And then they technically don't actually belong to any user session. There are some further long term ideas and more details …

So you mean that when you start an application inside a graphical session, systemd takes over it and forgets essential informations such as where it is actually running??

DaveMielke:
[quoted lines by Benjamin Berg on 2020/04/28 at 08:01 -0700]

>Yup, unfortunately we really have to. There is no way around it and it is safe
>because users can only have one graphical session by definition. Some people
>have not accepted that one yet though.

I myself am not a graphical user, so I might be wrong, but, as I understand it,
there's the multi-seat thing. To me, that means more than one graphical
session.

>The thing is that the processes of the graphical session end up being started
>by systemd. And then they technically don't actually belong to any user
>session. There are some further long term ideas and more details … but right
>now, guessing is definitely the right thing to do!

If there really is a right way to guess then shouldn't the place with the best
access to the data, i.e. (in this case) systemd, be providing a functino to do
it? I don't think it's right for a bunch of user code to implement that stuff
as it usually means not only that it might be done wrong but also that it won't
get fixed when there's a better way.

-- 

sthibaul:
> I myself am not a graphical user, so I might be wrong, but, as I understand it, there's the multi-seat thing. To me, that means more than one graphical session.

multi-seat is even something else: several sessions showing up *at the same time* because there are several screens & keyboard plugged on the computer. In Braille terms, it'd mean several Braille  devices (each of which used in combination with a screen and a keyboard) driven by several brltty daemons with several BrlAPI sockets, and we'd have to specify somewhere in the seat what socket Orca is supposed to connect to to get its output shown on the right Braille display.

In the multi-seat case, the same user should be able to log in in each seat and see things work properly indeed.

benzea:
> I myself am not a graphical user, so I might be wrong, but, as I understand it, there's the multi-seat thing. To me, that means more than one graphical session.

We are trying hard to kill that. It has been utterly broken since forever, but unfortunately people hold on to that idea because it kind of worked in the past, even though it is really bad.

If you run startx, then `sd_pid_get_session` *will* find it.

However, if you start a modern GNOME desktop (KDE is following, so is everyone else), then it will not work. What GNOME and others do is that they launch a systemd target/service which then starts the desktop and all applications run in the users scope. This launched target and all processes of the running session do not belong to a "session" technically. They "only" belong to the user from the point of view of systemd.

This is what my "session" looks like (systemd-cgls) on a standard F31:
```
-.slice
├─user.slice
│ └─user-1000.slice
│   ├─user@1000.service
│   │ ├─xdg-permission-store.service
│   │ │ └─2405659 /usr/libexec/xdg-permission-store
│   │  ... all user processes
│   └─session-14.scope
│     ├─2405377 gdm-session-worker [pam/gdm-password]
│     ├─2405396 /usr/bin/gnome-keyring-daemon --daemonize --login
│     ├─2405409 /usr/libexec/gdm-wayland-session /usr/bin/gnome-session
│     ├─2405412 /usr/libexec/gnome-session-binary
│     ├─2406745 /usr/bin/ssh-agent -D -a /run/user/1000/keyring/.ssh
│     └─2944438 gdm-session-worker [pam/gdm-password]
```

All other processes are launched as systemd service (or into a scope). They then are not part of the users `session-X.scope` breaking the session lookup from the PID.

benzea:
> I don't think it's right for a bunch of user code to implement that stuff as it usually means not only that it might be done wrong but also that it won't get fixed when there's a better way.

Yeah, it is a bit unfortunate that we are cargo-culting this code around. There is a DBus API which will magically do the right thing for you, however, that is harder to use than this.

We could probably fix it up in libsystemd do that magic there too. However, considering there are some unresolved issues around how multi-seat will work it might not be worthwhile right now.

Note that brltty should just assume `seat0`. Multi-seat is another can of worms that is probably best to just ignore here. It is a pretty niche thing to begin with, and I doubt someone will have 2 keyboards, 2 mice, >2 monitors **and** 2 braille displays on one machine to have two seats.

sthibaul:
> > I myself am not a graphical user, so I might be wrong, but, as I understand it, there's the multi-seat thing. To me, that means more than one graphical session.
> 
> We are trying hard to kill that.

To kill what? The multi-seat support, or the multiple graphical support by running on different VTs?

> What GNOME and others do is that they launch a systemd target/service which then starts the desktop and all applications run in the users scope.

That is fine.

> This launched target and all processes of the running session do not belong to a "session" technically. They "only" belong to the user from the point of view of systemd.

That is where this becomes unreasonable to me. Why shouldn't there be support for several sessions there?

> With regard to multi-login. The idea is that the desktop environment will expand across multiple logins in the future.

How will this look like exactly? Multiple VTs?

benzea:
> To kill what? The multi-seat support, or the multiple graphical support by running on different VTs?

Multiple separate graphical logins of one user. multi-seat is supposed to continue working.

i.e. one graphical session per user. This single graphical session my span multiple seats in the future (but not multiple VTs on one seat).

> How will this look like exactly? Multiple VTs?

No, one user cannot have multiple sessions across different VTs (assuming `seat0`). I don't think that seats other than `seat0` actually have VTs, but I might be wrong there.

You could, by starting a separate systemd user instance inside your session. But that is a huge hack and the only reason I entertain that idea is for development purposes. No user should ever have multiple graphical sessions … there are just too many horrible corner cases

DaveMielke:
[quoted lines by Benjamin Berg on 2020/04/28 at 09:43 -0700]

>We could probably fix it up in libsystemd do that magic there too. However,
>considering there are some unresolved issues around how multi-seat will work
>it might not be worthwhile right now.

If there are unresolved issues for you then wouldn't they also be unresolved
issues for us? So why not just provide the not yet perfect implementation and
then we'd automatically pick up the fixes as they're made?

-- 

sthibaul:
So, to summarize, `sd_pid_get_session` will work in the following cases:
- when logged in on a text VT (session type tty),
- when using startx from a VT (session type tty),
- when connected through ssh (session type tty),
and it will not work when logged in from gdm (session types x11 or wayland),
it would also not work when running from e.g. a cron or such.

And it happens that opening multiple gdm sessions for the same user would not be supported.

So in the end I guess doing this should be safe enough:
- if `sd_pid_get_session` works, use it (and ideally in the long run it'd work in all cases)
- otherwise (which means we are from gdm), look for sessions of the user, and find one which is not of type tty, normally there would be just one, which should actually be ours.

sthibaul:
(implementing that now)

sthibaul:
Ok, that seems to be working appropriately

benzea:
> So, to summarize, sd_pid_get_session will work in the following cases

Sorry, I was a bit inaccurate there. `startx` might exhibit it too, as it depends on how the applications are launched. i.e. directly by the "session" or indirectly by starting a `systemd` unit (which might also happen for DBus activated services).

Said differently, it depends on the graphical session. And GNOME will, by default, try to run using systemd (it currently falls back automatically if that doesn't work for some reason).

> So in the end I guess doing this should be safe enough:

Yep, that is the correct approach here. And, it is completely safe! The only corner case where it falls apart is simply not supportable currently.

DaveMielke:
[quoted lines by Benjamin Berg on 2020/04/29 at 01:36 -0700]

>Yep, that is the correct approach here. And, it is completely safe! The only
>corner case where it falls apart is simply not supportable currently.

And, for the ever curious (like myself), what is that corner case?

-- 

benzea:
> [quoted lines by Benjamin Berg on 2020/04/29 at 01:36 -0700]
> Yep, that is the correct approach here. And, it is completely safe! The only corner case where it falls apart is simply not supportable currently.
> And, for the ever curious (like myself), what is that corner case?
> […](#)

You first start a GNOME session (lets say normally through GDM) and then a GNOME one using startx. That way you can end up with two graphical sessions and the above code will give up.

With the similar code in mutter/GDM what will happen is that the guessing for applications in the GNOME sessions stops working suddenly. And then you will see weird stuff happening like PolicyKit refusing the session to set the brightness for the backlight.

sthibaul:
> Said differently, it depends on the graphical session. And GNOME will, by default, try to run using systemd (it currently falls back automatically if that doesn't work for some reason).

So with GNOME from startx, `sd_pid_get_session` will fail?

And thus with several GNOME from several startx it will fail, but as you say, GNOME won't support this case.

> > So in the end I guess doing this should be safe enough:
> 
> Yep, that is the correct approach here. And, it is completely safe! The only corner case where it falls apart is simply not supportable currently.

One case which will fail is GNOME from startx, then, since in that case `sd_pid_get_session` will fail, and `sd_uid_get_sessions` will only report a `tty` session, which we don't want to automagically try to attach to since we can't tell it appart from other `tty` sessions.

sthibaul:
> So with GNOME from startx, `sd_pid_get_session` will fail?

I tried startx gnome-session, and `sd_pid_get_session` worked. This is with gnome-session 3.36.0.

sthibaul:
> You first start a GNOME session (lets say normally through GDM) and then a non-GNOME one using startx. That way you can end up with two graphical sessions and the above code will give up.

No, when using startx you get a session of type `tty`.

sthibaul:
> > You first start a GNOME session (lets say normally through GDM) and then a non-GNOME one using startx. That way you can end up with two graphical sessions and the above code will give up.
> 
> No, when using startx you get a session of type `tty`.

And thus things work for both sessions: inside the startx-started session, `sd_pid_get_session` works, and inside the gdm-start session `sd_pid_get_session` will fail, but only the gdm-started session will have type `x11` or `wayland`.

sthibaul:
> when using startx you get a session of type `tty`.

(because the session was created when logging it on the VT)

benzea:
Oh, true. I had not thought about that corner case.

Is startx actually that relevant for us?

sthibaul:
> Is startx actually that relevant for us?

It is extremely relevant for a lot of blind users which only use the graphical session when they really have to (e.g. to browse a javascript-powered website).

benzea:
> It is extremely relevant for a lot of blind users who only use the graphical session when they really have to (e.g. to browse a javascript-powered website).

Ooops. For GNOME there is an easy workaround (run `gnome-session --builtin`), but that isn't really a solution …

This is an interesting problem, I don't have a nice solution right now. Well … various ideas for workarounds but nothing proper. A VT switch rather than startx would be one of the easier ones there …

sthibaul:
> For GNOME there is an easy workaround (run `gnome-session --builtin`),

Working around what, precisely?
It is completely fine to have sd_uid_get_session return the tty session opened before running startx, since that will properly report the VT.

> A VT switch rather than startx would be one of the easier ones there …

A VT switch from what to what?

sthibaul:
I mean, I don't see any issue any more, since you say that basically the gnome desktop won't support multiple graphical sessions. That is fine since there is usually only one gdm, and thus only one session that `sd_pid_get_session` can't return, but that we can get from `sd_uid_get_sessions`.

benzea:
If you run GNOME with `--builtin` it will not use systemd. And `sd_pid_get_session` will report the correct session.

With a VT switch I mean running the graphical session from e.g. GDM and using the normal tty logins alongside.

sthibaul:
> If you run GNOME with `--builtin` it will not use systemd. And `sd_pid_get_session` will report the correct session.

I'm lost.

Again, when running `startx gnome-session`, `sd_pid_get_session` already reports the correct session, which is the session created when the user logged in on the VT.

benzea:
> Again, when running startx gnome-session

Ohh, wow! I see now why!

 1. startx unsets `DBUS_SESSION_BUS_ADDRESS` (the systemd one).
 2. gnome-session detects the lack of the environment variable
 3. gnome-session runs `dbus-launch` internally
 4. gnome-session cannot talk to systemd (because it is a private dbus server instance)
 5. gnome-session falls back to legacy login and things work

Wow, I had not expected that chain of events!

sthibaul:
> > Again, when running startx gnome-session
> 
[...]
> Wow, I had not expected that chain of events!

Ok, so I guess this would be fragile.

If gnome "fixes" this into talking with systemd, a gnome session started from startx will also not be seen from `sd_pid_get_session`. That is fine with regards to a session opened in gdm since that is not supported.

The remaining issue I then see is starting as the same user a non-gnome session from gdm/xdm, and a gnome session from startx on another VT. Both will have `sd_pid_get_session` non-working, and `sd_uid_get_sessions` will report two non-tty sessions. But I'd say we don't want to try to fix this case inside libbrlapi.

benzea:
Well, I would argue it has always been fragile :)

And, I agree that a clean solution would be nice here. I don't have one though and it just adds a bit more to the overall complexity of the situation that needs to be solved … but I fear solutions will take quite a while.

sthibaul:
> I fear solutions will take quite a while.

At least for the time being this PR supports all usual cases, only not the case I mentioned (if gnome "fixes" using systemd, logging as non-gnome in gdm/xdm, and as gnome in startx, but then one can tell the user to use `gnome-session --builtin`).

PR #266, opened 2020-05-03
Install systemd units and udev rules correctly

Conan-Kudo:
This change makes it so that systemd units and udev rules, as well
as their associated helper scripts, are installed into the correct
directories as needed. We rely on the systemd and udev pkgconfig files
to provide the correct paths and install the helper scripts into
the correct directory for helper binaries.


Conan-Kudo:
@DaveMielke Why did you close this and not merge it?

DaveMielke:
Because I'm not sure it's the right thing to do. They used to be installed
there and someone else complained that that was wrong. So my solution now is to
put those wrappers in brltty's own lib directory. I'd think that that should be
okay. I'm willing to change this, but I'd need to be convinc I wish that
systemd would have a directory serving the same purpose as /lib/udev/ serves
for udev.

Conan-Kudo:
Most likely they complained because `/usr/libexec` wasn't standardized across all distributions. This changed in 2019 when both Debian and openSUSE changed to follow FHS 3.0 and now use the standard libexecdir. I also structured the patch so that if a distribution has an alternative location for `@libexecdir@`, they can pass it in through `./configure --libexecdir=` and everything works correctly.

As for the systemd unit file installation, the paths are now sourced from the systemd pkgconfig file, so they will _always_ be correct for the target distribution.

Conan-Kudo:
> Two things: What's wrong with brltty's wrappers being in brltty's own directory (/lib/brltty/)? I'm unaware of any standard that forbid's that. As I understand it, directories like /libexec/ are for programs which don't have their own place(s) to put things. Since brltty has its own place, why shouldn't it use that place? 

There's nothing wrong with it, but moving it from `/usr/lib(64)/brltty` to `/usr/libexec/brltty` will help when `noexec` is enforced on `/usr/lib` and `/usr/lib64`.

> This stanard is clearly evolving. Even if it's more or less settled down, now, that doesn't matter too much to me because I want those runninhg older distributions to still be able to build a modern brltty. Since we're now using a solution that works for newer as well as older distributions, why should we change it - especially since we aren't breaking any rules?

Well, the way I adapted it makes it so that distributions that haven't changed don't get negatively impacted, and those that have will get it doing the right thing.

DaveMielke:
[quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/05 at 09:03 -0800]

>There's nothing wrong with it, but moving it from `/usr/lib(64)/brltty` to
>`/usr/libexec/brltty` will help when `noexec` is enforced on `/usr/lib` and
>`/usr/lib64`.

What's the possibility of that happening? I, personally, don't see the
usefulness of denying the execution of scripts while, at the very same time,
alowing the execution of machine code (i.e. what's in .a and .so files).

>Well, the way I adapted it makes it so that distributions that haven't changed
>don't get negatively impacted, and those that have will get it doing the right
>thing.

The phrasee "the right thing" is subjective. For an unrelated example, aborting
a baby is considered "the right thing" to do by many but the deliberate,
unjustifiable murder of an unwanted person by those like myself.

Brltty is built by a lot of regular users, i.e. not just by distributions. One
of my goals is to make their lives as easy as possible since life for a braille
user without a working braille device is essentially as inconvenient as a
screenless system is to a sighted user.

Conan-Kudo:
> [quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/05 at 09:03 -0800]
> There's nothing wrong with it, but moving it from `/usr/lib(64)/brltty` to `/usr/libexec/brltty` will help when `noexec` is enforced on `/usr/lib` and `/usr/lib64`.
> What's the possibility of that happening? I, personally, don't see the usefulness of denying the execution of scripts while, at the very same time, alowing the execution of machine code (i.e. what's in .a and .so files).

`.a` and `.so` files do not require executable bit to work. So the possibility is quite high.

> Well, the way I adapted it makes it so that distributions that haven't changed don't get negatively impacted, and those that have will get it doing the right thing.
> The phrasee "the right thing" is subjective. For an unrelated example, aborting a baby is considered "the right thing" to do by many but the deliberate, unjustifiable murder of an unwanted person by those like myself. Brltty is built by a lot of regular users, i.e. not just by distributions. One of my goals is to make their lives as easy as possible since life for a braille user without a working braille device is essentially as inconvenient as a screenless system is to a sighted user.

Hyperbole aside there, there's no intent to break that for people who want to build and use it entirely from source.

DaveMielke:
[quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/05 at 12:30 -0800]

>`.a` and `.so` files do not require executable bit to work. So the possibility
>is quite high.

But that isn't my point. Why would it be done at all since code in .a and .so
files will still be executable? What would such a change actually gain?

>Hyperbole aside there, 

Not hyperbole at all. No one, even indirectly, will ever shame me into not
speaking to what really isn't right. I don't think anyone should throw around
phrases like "it's the right thing to do" in order to bully others into giving
in. Proper discussion about pros and cons is always welcome, but subjective
assertions aren't.

>there's no intent to break that for people who want to build and use it
>entirely from source.

Let's put it this way. When the udev (really the systemd) people stop making
[/usr]/lib[64]//udev/ be their default place for helper scripts then I'll bgin
to take it seriously. Of course, maybe that's already happened and I'm just not
aware of it.

Really, though, as long as code in .a and .so files remains executable then, to
me, it's nothing short of ridiculous to move to deny script execution. Let's
hear some actual pros and cons about that, rather than just subjectively
asserting that "it's the right thing to do".

Conan-Kudo:
> [quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/05 at 12:30 -0800]
> `.a` and `.so` files do not require executable bit to work. So the possibility is quite high.
> But that isn't my point. Why would it be done at all since code in .a and .so files will still be executable? What would such a change actually gain?

The gain is that it closes an attack vector by the means of injecting content into `/usr/lib(64)` that could be directly executed when that is not expected. The code inside _cannot_ be directly executed, only sourced by trusted binaries in `/usr/(s)bin` or `/usr/libexec`. It's about constraining paths to roles.

> Hyperbole aside there,
> Not hyperbole at all. No one, even indirectly, will ever shame me into not speaking to what really isn't right. I don't think anyone should throw around phrases like "it's the right thing to do" in order to bully others into giving in. Proper discussion about pros and cons is always welcome, but subjective assertions aren't.

In this instance, "doing the right thing" is following the conventions that most (if not all) Autotools scripts would do for this. It is not my intent to "bully" anyone into anything. I originally made this patch to fix the build and install on Fedora and openSUSE in a cleaner way than how it was done downstream for the benefit for all consumers of brltty.

> there's no intent to break that for people who want to build and use it entirely from source.
> Let's put it this way. When the udev (really the systemd) people stop making [/usr]/lib[64]//udev/ be their default place for helper scripts then I'll bgin to take it seriously. Of course, maybe that's already happened and I'm just not aware of it. Really, though, as long as code in .a and .so files remains executable then, to me, it's nothing short of ridiculous to move to deny script execution. Let's hear some actual pros and cons about that, rather than just subjectively asserting that "it's the right thing to do".

Udev already does not install stuff into `/usr/lib64/udev`. Helper scripts have long since been recommended to _not_ be installed in the udev directories, either. As for `.a` and `.so` code files, the long-term goal is to make it so that those files are compiled as _non-executable_ content. There's work in package build infrastructure to already start making this happen. But otherwise I'm not sure what you're saying here...

DaveMielke:
[quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/05 at 13:09 -0800]

>The gain is that it closes an attack vector by the means of injecting content
>into `/usr/lib(64)` that could be directly executed when that is not expected.

Sure, I get that. Over-writing a script like /lib/brltty/systemd-wrapper could
be done,wouldn't be nice, and would be very naïvely executed.

I guess the critical word, here, is "directly".

>The code inside _cannot_ be directly executed, only sourced by trusted
>binaries in `/usr/(s)bin` or `/usr/libexec`.

What makes those directly executable binaries any more trustworthy. Any of them
could be overwritten just as easily as any script under /lib/. To me, binaries
aren't somehow more trustworthy than scripts.

>It's about constraining paths to roles.

But, perhaps ignorantly, I don't see what the role difference is? Both are
software that users don't directly use but that are used to augment the
service(s) of the binaries that users do invoke.

>In this instance, "doing the right thing" is following the conventions that
>most (if not all) Autotools scripts would do for this. 

As with everything, it's not easy for most of us to know what's become a truly
used and worthwhile convention and what's just a trend. I could, for exmple,
point to the (now rather ancient) devfs fad. I, for one, am sure glad that udev
replaced it. For a while, though, we were all effectively told to wholly
embrace it. I'll always end up doing "the right thing", but I'm cautious and do
need convincing.

>It is not my intent to "bully" anyone into anything.

Yes, I know that. The general problem, though, is that innocent peolpe often
get caught up in systems and processes that do.

Please understand my position. I get a PR from someone I don't know that does
something rather invasive. I have a project who's integrity I "work overtime"
to protect. I have a project that's highly trusted. What am I supposed to do?

What I did decide to do what just close the PR. If I then would get no response
then no big deal. If I then get a response then we can have a discussion. The
latter is what happened, and, for that, I'm grateful.

The only thing I regret is that this discussion didn't happen before the PR was
made. I think that'd have been a much better approach.

And, as I think I mentioned earlier but which you wouldn't have known at the
time, we did indeed have at least the systemd wrapper in /libexec/ (I think the
udev wrapper was in /lib/udev/). We were told that that was no good (I forget
the reason). My assumption was that, probably, some distributiion(s) just
didn't support doing things that way. You must understand that this, too,
factored into my first just not dealiong with, and then eventually closing, the
PR.

>I originally made this patch to fix the build and install on Fedora and
>openSUSE in a cleaner way than how it was done downstream 

Hey! Fedora may not know it, but it's my friend. :-) It's the distribution that
I use.

Do you, by the way, work for one of them?

>for the benefit for all consumers of brltty.

I understand. 

>Udev already does not install stuff into `/usr/lib64/udev`. Helper scripts
>have long since been recommended to _not_ be installed in the udev
>directories, either. 

Okay, I didn't know that. I guess, when I look in /lib/udev/, I'm just then
seeing the list of projects that are either resisting or not caring. :-(

>As for `.a` and `.so` code files, the long-term goal is to make it so that
>those files are compiled as _non-executable_ content. There's work in package
>build infrastructure to already start making this happen. 

That's actually rather cool, though I don't understand how it would actually
improve things. Either way, the nature of a lib is to provide executable
functionality to a binary. Whether or not it's executable wouldn't change that.
For exmple, even a .a archive doesn't need to be executable. It's members are
simply loaded and the machine code contaioned therein is still blindly
executed.

>But otherwise I'm not sure what you're saying here...

Maybe it's because you know far more about these things than I do (which is
entirely possible). I have no problem admitting that I don't know something,
and would much rather have a discussion than remain ignorant.

So, this discussion is actually very timely. We released 6.2 a few weeks ago,
and, for a number of reasons, I'm intending to release 6.2.1 in a couple of
weeks. If we include this then I'll call it 6.3 instead.

It's probably too late for PR266 to apply cleanly, but I don't mind looking at
what it does and reworking it.

DaveMielke:
In /libexec/, should a package have it's own subdirectory or should a prefix
like brltty- be used?

Conan-Kudo:
Subdirectory is better.

DaveMielke:
[quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/05 at 14:39 -0800]

>Subdirectory is better.

Thought so. Done. Have a look at the libexec branch. I added
--with-scripts-directory, and each (systemd, udev) make file only needed a
one-line change. Doing make (un)install for both works (but you should check it
out). Just need to fight with autoconf variables so that ${execprefix} doesn't
show up in the systemd service and udev rules paths to the scripts.

DaveMielke:
Okay. I think I'm done. Please check out the libexec branch. If you've pulled
already then you'll need to use -f as I used --amend. If you're happy with it
then I'll merge.

DaveMielke:
Is the libexec branch okay? Shall I merge it?

Conan-Kudo:
@DaveMielke `libexec` branch looks good to me. After that's merged, I can rebase the leftover bits of this patch for requesting the correct unit and udev rules paths using pkgconfig.

DaveMielke:
Okay, it's merged now.

Conan-Kudo:
I'll rebase my patch ASAP, and hopefully have something for you to review in the next few days.

DaveMielke:
I see, now, what you're doing with pkgconfig. I'll add that and create a branch
for you to check out. It needs to be done soon since I've already announced 6.3
for a week and a half from now.

I wonder if brltty should have a pkgconfig file.

Conan-Kudo:
pkgconfig files are used for helping build scripts detect where headers and development libraries are, and in some cases, install locations for extensions. I don't know if brltty would actually qualify for any of those scenarios to warrant its own pkgconfig file.

DaveMielke:
Have a look at the pkgconfig branch. Anything missing? Two questions:

Any reason that the AC_SUBST() lines hard-code /usr/lib? Shouldn't they be
using ${libdir}?

You used a paradigm that I've never used before, i.e. assigning a specific
value to a variable, and then AC_SUBSWTing it later with what I guess is a
deault value. Is that correct? Does AC_SUBST() not assign its second argument
if the variable already has a value?

Yes, brltty has a rough equivalent of a ppkgconfig file in the form of
/usr/bin/brltty-config. It's embedded (via .) into other scripts to get
build-time values.

Conan-Kudo:
@DaveMielke I'm a bit naive with autotools scripting. I was trying to follow what was done elsewhere and try to make sure that it would pick up a default value correctly while letting you override it if need be.

DaveMielke:
[quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/07 at 09:43 -0800]

>@DaveMielke I'm a bit naive with autotools scripting. I was trying to follow what was done elsewhere and try to make sure that it would pick up a default value correctly while letting you override it if need be.

Okay, I'll fix it to default to ${libdir} instead of to the hard-coded
/usr/lib. As long as I know the intent, I know what to do.

Conan-Kudo:
> Any reason that the AC_SUBST() lines hard-code /usr/lib? Shouldn't they be using ${libdir}?

For reasons I'm not certain of, systemd puts vendor provided config files in `/usr/lib` (instead of `/usr/share`) while allowing user/admin overrides to live in `/etc`.

Conan-Kudo:
> Okay, I'll fix it to default to ${libdir} instead of to the hard-coded /usr/lib. As long as I know the intent, I know what to do.

It was definitely intended to use `/usr/lib`, as that's a hardcoded systemd thing.

DaveMielke:
Okay, please check out the latest pkgconfig branch. Again, if you've fetched
already then you'll need to use -f.

DaveMielke:
If I create a pkgconfig file for brltty, what's the rule? I have .pc files in
both /usr/share/pkgconfig/ (e.g. udev.pc) and in /usr/lib64/pkgconfig/ (e.g.
libudev.pc).

Conan-Kudo:
if the pc file describes architecture specific stuff (like archful file paths like `/usr/lib64`), then it goes in `/usr/lib64/pkgconfig`, otherwise it goes into `/usr/share/pkgconfig`.

DaveMielke:
How does the pkgconfig branch look? Anything else to do? Yes, I know I still
need to update the rpm spec file.

I've attached the (new) brltty.pc.in and brltty.pc files. Do they look okay to
you?

###############################################################################
# BRLTTY - A background process providing access to the console screen (when in
#          text mode) for a blind person using a refreshable braille display.
#
# Copyright (C) 1995-2021 by The BRLTTY Developers.
#
# BRLTTY comes with ABSOLUTELY NO WARRANTY.
#
# This is free software, placed under the terms of the
# GNU Lesser General Public License, as published by the Free Software
# Foundation; either version 2.1 of the License, or (at your option) any
# later version. Please see the file LICENSE-LGPL for details.
#
# Web Page: http://brltty.app/
#
# This software is maintained by Dave Mielke <dave@mielke.cc>.
###############################################################################

Name: @PACKAGE_TARNAME@
Version: @PACKAGE_VERSION@
URL: @PACKAGE_URL@
Description: Braille Device Daemon

package_name=@PACKAGE_TARNAME@
package_title=@PACKAGE_NAME@
package_version=@PACKAGE_VERSION@
package_url=@PACKAGE_URL@
bugreport_email=@PACKAGE_BUGREPORT@

prefix=@prefix@
exec_prefix=@exec_prefix@
datarootdir=@datarootdir@

sysconfdir=@sysconfdir@

execute_root=@execute_root@
program_directory=@program_directory@
drivers_directory=@drivers_directory@
scripts_directory=@scripts_directory@
tables_directory=@tables_directory@
updatable_directory=@updatable_directory@
writable_directory=@writable_directory@
manpage_directory=@manpage_directory@
include_directory=@include_directory@

braille_parameters=@braille_parameters@
speech_parameters=@speech_parameters@
screen_parameters=@screen_parameters@
privilege_parameters=@privilege_parameters@

api_version=@api_version@
api_release=@api_release@
api_parameters=@api_parameters@
api_sockets_directory=@BRLAPI_SOCKETPATH@
api_authkey_file=@BRLAPI_ETCDIR@/@api_authkeyfile@


###############################################################################
# BRLTTY - A background process providing access to the console screen (when in
#          text mode) for a blind person using a refreshable braille display.
#
# Copyright (C) 1995-2021 by The BRLTTY Developers.
#
# BRLTTY comes with ABSOLUTELY NO WARRANTY.
#
# This is free software, placed under the terms of the
# GNU Lesser General Public License, as published by the Free Software
# Foundation; either version 2.1 of the License, or (at your option) any
# later version. Please see the file LICENSE-LGPL for details.
#
# Web Page: http://brltty.app/
#
# This software is maintained by Dave Mielke <dave@mielke.cc>.
###############################################################################

Name: brltty
Version: 6.2
URL: http://brltty.app/
Description: Braille Device Daemon

package_name=brltty
package_title=BRLTTY
package_version=6.2
package_url=http://brltty.app/
bugreport_email=BRLTTY@brltty.app

prefix=
exec_prefix=${prefix}
datarootdir=${prefix}/share

sysconfdir=${prefix}/etc

execute_root=/brltty/REV-BRLTTY-6.2-50-g34d3c1f8M
program_directory=${exec_prefix}/bin
drivers_directory=${exec_prefix}/lib/brltty
scripts_directory=${exec_prefix}/libexec/brltty
tables_directory=${prefix}/etc/brltty
updatable_directory=${prefix}/var/lib/brltty
writable_directory=/run/brltty
manpage_directory=${datarootdir}/man
include_directory=${prefix}/include/brltty

braille_parameters=
speech_parameters=
screen_parameters=
privilege_parameters=

api_version=0.8
api_release=0.8.1
api_parameters=
api_sockets_directory=/var/lib/BrlAPI
api_authkey_file=/etc/brlapi.key

Conan-Kudo:
Wouldn't drivers be installed into `${libdir}/brltty`?

Conan-Kudo:
But otherwise, looks good to me.

DaveMielke:
[quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/07 at 12:49 -0800]

>Wouldn't drivers be installed into `${libdir}/brltty`?

They are. Perhaps I should change the autoconf variable that's used to more
clearly show that.

This brings up another qustion. autoconf is setting libdir to lib, not to
lib64. Is that expected?

DaveMielke:
[quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/07 at 12:54 -0800]

>But otherwise, looks good to me.

The .pc file, the pkgconfig branch, or both.

-- 

Conan-Kudo:
> [quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/07 at 12:49 -0800]
> Wouldn't drivers be installed into `${libdir}/brltty`?
> They are. Perhaps I should change the autoconf variable that's used to more clearly show that. This brings up another qustion. autoconf is setting libdir to lib, not to lib64. Is that expected?

If it was a 32-bit system, yes. If it's a 64-bit system, no.

Conan-Kudo:
> [quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/07 at 12:54 -0800]
> But otherwise, looks good to me.
> The .pc file, the pkgconfig branch, or both.
> […](#)
> -- I believe the Bible to be the very Word of God: http://Mielke.cc/bible/ Dave Mielke | 2213 Fox Crescent | WebHome: http://Mielke.cc/ EMail: Dave@Mielke.cc | Ottawa, Ontario | Twitter: @Dave_Mielke Phone: +1 613 726 0014 | Canada K2A 1H7 |

The .pc file. Hadn't had a chance to look at the branch yet.

DaveMielke:
[quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/07 at 14:24 -0800]

>If it was a 32-bit system, yes. If it's a 64-bit system, no.

Maybe Fedora configures with --libdir=/usr/lib64, then.

Looking at your changes to the rpm spec file: Are _unitdir and _sysusersdir
preset? Is there also a _tmpfilesdir variable?

Conan-Kudo:
> [quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/07 at 14:24 -0800]
>> If it was a 32-bit system, yes. If it's a 64-bit system, no.
> Maybe Fedora configures with --libdir=/usr/lib64, then. 

We do.

> Looking at your changes to the rpm spec file: Are _unitdir and _sysusersdir preset? Is there also a _tmpfilesdir variable?

Yes, there is. These are all provided by the `systemd-rpm-macros` package (it used to be part of `systemd` and is now split out), so you need to `BuildRequires: systemd-rpm-macros`.

DaveMielke:
The rpmspec branch is now ready for you to have a look at. Is tehre anything
else that can be done to the spec file to make it easier?

DaveMielke:
I hope you've had an opportunity to look at the pkgconfig and rmpsepc branches.

I'm also wondering if you'd like the pkgconfig variables actiondir and
policydir from polkit-gobject-1.pc to be used.

Any other improvements to the rpm that you can think of?

Conan-Kudo:
pkgconfig branch looks good, looking at rpmspec branch now

Conan-Kudo:
rpmspec looks good so far

Conan-Kudo:
> I'm also wondering if you'd like the pkgconfig variables actiondir and policydir from polkit-gobject-1.pc to be used.

Hah, that probably makes sense. I didn't even know those existed! 😅

DaveMielke:
[quoted lines by Neal Gompa (ニール・ゴンパ) on 2021/01/08 at 11:51 -0800]

>> I'm also wondering if you'd like the pkgconfig variables actiondir and policydir from polkit-gobject-1.pc to be used.
>
>Hah, that probably makes sense. I didn't even know those existed! 😅 

See brltty's Authorization/Polkit/ subdirectory.

Conan-Kudo:
Ahh, yeah. So those probably make sense to retrieve the paths from the polkit pkgconfig files.

DaveMielke:
I forgot to ask if there are rpm variables for polkit's actiondir and
policydir. I yes, do we need to have anotehr BuildRequires for them?

Conan-Kudo:
Doesn't appear to be any rpm macros for polkit...

DaveMielke:
Unfortunately, both polkit's actiondir and policydir variables point to the
same directory (makes sense - they're essentially synonyms), but there's no
variable for the rules.d directory. :-(

DaveMielke:
Okay. The pkgconfig branch is merged, and the policydir polkit variable is
being used. The code to support a variable for polkit's rules.d directory is
in, and, for now, it assumes the variable name ruledir and (since it has no
value) uses the correct default value.

So is the rpmspec branch okay, and what else would be useful to add to the
spec file?

Conan-Kudo:
The `rpmspec` branch looks okay to me. I can't think of anything else to add right now...

DaveMielke:
The current rpm spec file just builds the BrlAPI Python bindings with Python3.
Is there any need to still support Python2? As far as I know, the only major
app that uses it is Orca, which I believe has been using Python3 for quite some
time, now.

Conan-Kudo:
No need for Python 2.

DaveMielke:
Are there rpm variables for the .pc file install directories?

The current plan is to release brltty-6.3 on Monday, January 18.

Conan-Kudo:
Nope.

Just use `%{_libdir}/pkgconfig` or `%{_datadir}/pkgconfig` depending on where it's going to be installed.

DaveMielke:
We have these lines near the start of our spec file:

   %define _bindir /bin
   %define _sbindir /sbin
   %define _libdir /lib
   %define _datadir /usr/share
    36 %define _sysconfdir /etc

I no longer remember why they're there. Do these variables/macros already
exist? Maybe they have different names. If they do already exist then why don't
these assignments seem to interfere with higher level processes which might be
assigning different values (e.g. _libdir to lib64)? Should these lines be
removed?

Conan-Kudo:
You shouldn't need those anymore, as they're defined by RPM itself already.

DaveMielke:
If you have the time, would you mind building an rpm with the latest code, and
maybe even install it, to check if everything really is good from your
perspective?

Conan-Kudo:
I don't think I'll have time this week due to work constraints, but I can try afterward.

PR #269, opened 2020-06-07
fix input_event time related compile fail

hthiery:
./system_linux.c: In function 'writeInputEvent':
./system_linux.c:962:23: error: 'struct input_event' has no member named 'time'; did     you mean 'type'?
   gettimeofday(&event.time, NULL);
                       ^~~~
                       type

Signed-off-by: Heiko Thiery <heiko.thiery@gmail.com>


hthiery:
The issue is observed with linux 5.7 headers on arm 32bit with MUSL c lib.

PR #270, opened 2020-06-11
brlapi: Add missing byteorder conversion when server notifies a param…

sthibaul:
…eter update

### DIFF
--- a/Programs/brlapi_server.c
+++ b/Programs/brlapi_server.c
@@ -2306,6 +2306,7 @@ static void __handleParamUpdate(Connection *dest, brlapi_param_t param, brlapi_p
   paramValue->subparam_hi = htonl(subparam >> 32);
   paramValue->subparam_lo = htonl(subparam & 0xfffffffful);
   memcpy(p, data, size);
+  _brlapi_htonParameter(param, paramValue, size);
   size += sizeof(flags) + sizeof(param) + sizeof(subparam);
   if (!(flags & BRLAPI_PARAMF_GLOBAL)) {
     sendConnectionParamUpdate(dest,param,subparam,flags,paramValue,size);


DaveMielke:
Doesn't there also need to be a corresponding ntoh in the client?

sthibaul:
> Doesn't there also need to be a corresponding ntoh in the client?

Since I had tested this successfully I assumed that it was already there, but I can't find it in `brlapi__doWaitForPacket`'s call to `callback->func`, so now I'm unsure

DaveMielke:
[quoted lines by Samuel Thibault on 2020/06/11 at 17:03 -0700]

>Since I had tested this successfully I assumed that it was already there, but I can't find it in `brlapi__doWaitForPacket`'s call to `callback->func`, so now I'm unsure

I think the patch should be something like this. I'm not sure how the correct
length should be specified, though (rlen is undefined).

diff --git a/Programs/brlapi_client.c b/Programs/brlapi_client.c
index 4242668f6..4628d5c15 100644
--- a/Programs/brlapi_client.c
+++ b/Programs/brlapi_client.c
@@ -486,6 +486,7 @@ static ssize_t brlapi__doWaitForPacket(brlapi_handle_t *handle, brlapi_packetTyp
     brlapi_param_flags_t flags = ntohl(value->flags);
     brlapi_param_t param = ntohl(value->param);
     brlapi_param_subparam_t subparam = ((brlapi_param_subparam_t)ntohl(value->subparam_hi) << 32) | ntohl(value->subparam_lo);
+    _brlapi_ntohParameter(param, value, rlen);
     pthread_mutex_lock(&handle->callbacks_mutex);
     for(handle->nextCallback = handle->parameterCallbacks;
 	handle->nextCallback; ) {

PR #279, opened 2020-11-02
xbrlapi: Also print some debugging when getting clipbord content from X

sthibaul:


DaveMielke:
Rebased so mightn't be closed.

sthibaul:
Closing by hand indeed.

PR #281, opened 2020-11-03
Fix relations with X

sthibaul:
since program_directory may be "${exec_prefix}/bin" and exec_prefix may be
${prefix}.


DaveMielke:
Added HAVE_PTHREAD_ATFORK to config.h.in, and also some #endif comments.

PR #283, opened 2020-12-01
tt, vs: Fix link

sthibaul:
Both tt and vs were providing a keyToCommand function.


DaveMielke:
Rebased. Please close.

PR #285, opened 2020-12-01
polkit: Fix installation when building out of source

sthibaul:


DaveMielke:
Could you change them to $(SRC_DIR). Then you don't need to specify the
Authroization/Polkit part. $(SRC_DIR) means the same directory but in the
source tree.

sthibaul:
done so!

PR #298, opened 2020-12-30
XWindow: Add backspace key

sthibaul:


DaveMielke:
Rebased. Please close. Should've checked first. :-(

PR #307, opened 2021-01-23
[setup scripts] check for correct pulse group, check for bluetooth group

humenda:
The setup scripts for running BRLTTY as non-root user now check for the
correct pulse group on Debian and offer the possibility to also add the
user to the bluetooth group.

### DIFF
--- a/brltty-mkuser
+++ b/brltty-mkuser
@@ -89,6 +89,7 @@ addProgramOption k flag noKeyboards "don't allow keyboard monitoring"
 addProgramOption l string.path loginShell "set/change the login shell"
 addProgramOption p flag noPulse "don't allow playing sound via the Pulse Audio server"
 addProgramOption s flag noSerial "don't allow access to serial devices"
+addProgramOption t flag noBluetooth "don't allow access to bluetooth devices"
 addProgramOption u flag noUSB "don't allow access to USB devices"
 addProgramOption E flag allowChanges "allow changes to an existing user"
 addProgramOption G flag noGroups "don't set/change the supplementary group list"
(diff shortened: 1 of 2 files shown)


DaveMielke:
Testing for the pulse group isn't the right way. I've merged the PR but we need
to find a better answer.

The pulse group exists as the group that the pulse audio server should be
running as, and, as such, exists everywhere but isn't a group that any user
(other than pulse audio itself) should belong to. pulse-access is a
well-defined group within the pulse audio documentation so it's difficult for
me to beleave that it isn't the right group. The error, if there is one, is
likely that the distribution that you're using hasn't defined it.

Regarding the bluetooth group: Is this a group that the distribution you're
using has defined as a way to control the ability to use Bluetooth? I haven't
heard of that before. I think you're using Debian, and no "Debian users have
ever raised this as a concern before.

humenda:
>The pulse group exists as the group that the pulse audio server should be
>running as, and, as such, exists everywhere but isn't a group that any user
>(other than pulse audio itself) should belong to. pulse-access is a
>well-defined group within the pulse audio documentation so it's difficult for
>me to beleave that it isn't the right group. The error, if there is one, is
>likely that the distribution that you're using hasn't defined it.

Ok, I was not aware of this. I have only found pulse-access mentioned together
with Pulse as system daemon, but I suppose you are right. I saw the pulse
group and thought it to be the right one. Please revert.

>Regarding the bluetooth group: Is this a group that the distribution you're
>using has defined as a way to control the ability to use Bluetooth? I haven't
>heard of that before. I think you're using Debian, and no "Debian users have
>ever raised this as a concern before.

I have been debugging BRLTTY "as user" for a while and had bluetooth
connectivity issues. I have retried it and it seems that the bluetooth group
is not required. My user used to be part of it so I expected this to be a
common pattern. So sorry, but this can be reverted then as well.

The scripts from the BRLTTY source link to `#!/bin/sh`, but use Bash syntax.
While a lot of distributions use sh and bash as synonyms, they are
unfortunately not. Maybe you could spell out bash in the interpreter line so
that it also works on Debian derivatives.
Thanks

DaveMielke:
Yes, I'm aware but often forget, that sh is dash on Debian systems. On mine, sh
is bash so I don't notice it. I'll try to do a better job of remembering this
in the future.

PR #340, opened 2021-05-11
brlapi python: Update brlapi_param_properties_t fields

sthibaul:
The C code generated by cython will use the bitfields naturally.


DaveMielke:
The bit fields don't need (or can't be) annotated with :1? Does Cython just know anyway?

PR #342, opened 2021-06-21
French l10n: various fixes

aaclause:
Just some spelling/typo fixes for French localization. Thanks.


shindere:
Thanks a lot, André!!

DaveMielke:
If you're wondering why I ddn'≥ just directly use the PR, it's because of my conversion from building with ant to building with gradle. A cosequence of that is that the res/ directory moved from Android/Application/ to Android/Gradle/app/src/main/, so it'd have been a lot of work to try to fight with the PR. I just made a diff, therefore, and applied it to the new structure.

aaclause:
@DaveMielke no problem! Thanks for merging this. :)

PR #344, opened 2021-08-22
Windows: Factorize NSIS scripts

sthibaul:


DaveMielke:
The variant for libusb-1.0.nsi is libusb. Is that correct, or should I change it to libusb-1.0?

sthibaul:
Ah, that was a typo when I reworked by factorization. It should be changed to libusb-1.0 indeed
(though it's really not used much, apparently only for the installer title)

PR #349, opened 2021-10-09
Revert "Use fchmod (not chmod) to adjust the BrlAPI socket permissions. (dm)"

sthibaul:
This reverts commit 81cc72e023f6ca0fe6952d9b266ed937b38d28bc.

fstat on a unix socket does not actually get the socket inode status, so
that cannot be used for fixing the permissions.

### DIFF
--- a/Drivers/Screen/AtSpi2/a2_screen.c
+++ b/Drivers/Screen/AtSpi2/a2_screen.c
@@ -1454,7 +1454,10 @@ addWatches (void) {
   };
 
   for (const WatchEntry *watch=watchTable; watch->message; watch+=1) {
-    if (!addWatch(watch->message, watch->event)) return 0;
+    if (!addWatch(watch->message, watch->event)) {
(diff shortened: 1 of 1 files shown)


sthibaul:
> I'm extremely reluctant to revert that commit because it resolves a security exploit.
Yeah, I understand that.

> If fstat() really returns incorrect data
It does, I tried this 
[test.c.txt](https://github.com/brltty/brltty/files/7316607/test.c.txt) (I'm still amazed that github doesn't allow attaching a .c source file...)
it displays 140777 with fstat, and 140755 with stat

> verify that it returns the same st_dev and st_ino numbers
The thing is: they don't.

```
8 19380519 140777
fd02 28340293 140755
```
which is most probably the same reason why it's not the same st_mod.

DaveMielke:
ls -l /proc/<pid>/fd/<fd> - or readlink() on that path - should show which object fstat() is referring to. I'd much rather understand the problem than assume that something isn't working right.

PR #353, opened 2022-01-13
configure.ac: fix LD, STRIP and RANLIB hardcoding

10ne1:
When building the project with LLVM/Clang it is desirable to
be able to use the LLD linker as well as llvm-strip/ranlib
so replace the path-based hardcoded tool name search with a
more flexible AC_CHECK_TOOL search.

Signed-off-by: Adrian Ratiu <adrian.ratiu@collabora.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -404,9 +404,9 @@ else
 fi
 AC_SUBST([LIBCXXFLAGS])
 
-AC_PATH_TOOL([LD], [ld])
-AC_PATH_TOOL([STRIP], [strip])
-AC_PATH_TOOL([RANLIB], [ranlib])
+AC_CHECK_TOOL([LD], [ld])
+AC_CHECK_TOOL([STRIP], [strip])
+AC_CHECK_TOOL([RANLIB], [ranlib])
 
 AC_PROG_AWK
 AC_PROG_LN_S


10ne1:
Friendly ping? Anyone? cc @DaveMielke @mlang (Sorry I got these names from the readme).

DaveMielke:
I  already answered on the list (as someone asked there) that I'm not accepting this PR. It reverts building brltty to be single- (rather than cross-) platform, and it also assumes that certain environment variables are already defined.

I also don't know what this issue has to do with building brltty for Chrome OS. I know that Google has been building brltty for that platform for a long time and this isn't an issue for them.

Anyway, I think the goal is to not look for a tool if its associated environment variable is already set. If that's correct then please give me a list of all the ones you care about and I'll have a look at coding the change properly.

10ne1:
@DaveMielke Thanks for your offer to help, I did not see the response on the ML. Can we please continue the discussion here?

> I also don't know what this issue has to do with building brltty for Chrome OS. I know that Google has been building brltty for that platform for a long time and this isn't an issue for them.

Actually it is. I am working with Google / ChromeOS / ChromiumOS and Gentoo for this. The idea is to block and fix mixed compiler toolchain usage. There are packages in CrOS & Gentoo which build using GCC/binutils tools even though the build should be using only Clang/LLVM. Brltty is one of them.

[Here is a CrOS CL](https://chromium-review.googlesource.com/c/chromiumos/overlays/chromiumos-overlay/+/3105726/) which adds some blocking logic which makes building brltty fail. This is only used to detect offenders for now, but we'd like to merge it after fixing the offenders.

> Anyway, I think the goal is to not look for a tool if its associated environment variable is already set. If that's correct then please give me a list of all the ones you care about and I'll have a look at coding the change properly.

That is correct. The full list we're currently blocking in CrOS is `gcc ld nm as objcopy strip ranlib`. In the end we'd like to block all GNU tools for which an LLVM counterpart exists. From what I can see only ld, strip and ranlib are used in brltty.

Thanks again and please ping if you need any more context, if you want me to test something in CrOS, or something else.

10ne1:
Hi @DaveMielke Gentle ping, any progress or ideas on the proper solution you mentioned? Thanks!

10ne1:
@DaveMielke Yes, I can confirm just a AC_PATH_TOOL  -> AC_CHECK_TOOL change works. How do you want to continue? Can you apply this change to the master branch or want me to update this PR?

DaveMielke:
Up to you. I can do it, but, if you'd like credit for the fix, update the PR. Either way is fine by me.

10ne1:
@DaveMielke I updated the PR and added a suggested-by tag for you. Thanks!

DaveMielke:
Merged. Now you must tell the ever-curious me what 10ne1 means. :-)

10ne1:
It's just the diminutive form of my name (Ionel), written as 10ne1. :)

https://www.behindthename.com/name/ionel

PR #362, opened 2022-03-18
Handle Debian's default posix_prefix sysconfig layout

stefanor:
Fixes Debian's bug: https://bugs.debian.org/1007927

Debian adds a custom sysconfig scheme to system python installs, "posix_local". This is the default scheme, and it redirects local users' Python module installs to /usr/local even though Python is installed with a /usr prefix. Both are on Debian's python's sys.path module search path.

Previously this custom sysconfig scheme was specified in distutils.sysconfig, but not sysconfig itself. As distutils is being deprecated, the custom scheme is now specified in sysconfig, since Debian's Python 3.10 (3.10.2-4).

As far as I can see, pythoncmd is only used in detecting the installed system python, not determining where the built module will be installed, so this patch is sufficient. To control where the module is installed, the approach would be to pass the selected prefix to sysconfig, when querying the install path.

### DIFF
--- a/Tools/pythoncmd
+++ b/Tools/pythoncmd
@@ -50,6 +50,9 @@ try:
     scheme = sysconfig.get_default_scheme()
   else:
     scheme = sysconfig._get_default_scheme()
+  if scheme == 'posix_local':
+    # Debian's default scheme installs to /usr/local/ but we want to find headers in /usr/
+    scheme = 'posix_prefix'
 
   def getIncludeDirectory():
     return sysconfig.get_path("include", scheme)
--- a/Tools/pythoncmd
+++ b/Tools/pythoncmd
@@ -46,15 +46,19 @@ except IndexError:
 try:
   import sysconfig
 # putProgramMessage("sysconfig")
+  if hasattr(sysconfig, 'get_default_scheme'):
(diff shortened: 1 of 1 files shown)


DaveMielke:
The hasattr else clause says: scheme = sysconfig._get_default_scheme()
How far back is the function call _get_default_scheme() (i.e. with the underscore prefix) supported, i.e. when was it introduced?
My goal is to not break builds against older Python releases.

stefanor:
> How far back is the function call _get_default_scheme() (i.e. with the underscore prefix) supported, i.e. when was it introduced?

2.7.4

DaveMielke:
I have 2.7.18, so it of course works. I've merged your fix. If it ever fails with an really early 2.7 then we can always add a try clause.

I'm hoping that it was part of the very first release of sysconfig, but I'm not so sure that's true. Do you know?

stefanor:
> I'm hoping that it was part of the very first release of sysconfig, but I'm not so sure that's true. Do you know?

Yes, it was. I hadn't even realised that.

stefanor:
And of course, thanks!

DaveMielke:
I appreciate your fix. I don't use a Debian-based distribution so would never have even noticed the problem.

PR #371, opened 2022-04-03
Atspi2: Also convert \r

sthibaul:
It seems some clipboard contents get \r instead of \n, so convert them as
well.


DaveMielke:
When this is working, do you think that bit of code should be moved into the core so other screen drivers c an call it?

sthibaul:
I was wondering. Normally ttys do interpret control-j / control-m correctly, and it's really at-spi2 which is special in that it also emits characters to graphical applications which are not a tty.

sthibaul:
Actually the windows screen driver may make use of it, it's the same situation.

PR #388, opened 2022-06-14
Fix QBrailleXL unrecognizable in the second connection

proguest1942:
The "name" pointer will be free in line#655. This will make device->name return a memory position which already be free. This change make the free() always free the duplicated name instead of the name in memory.


DaveMielke:
Thanks. This problem means that 6.5.1 won't be too far off.

It wasn't the best fix, though. The right one, now committed on master, fixes the actual underlying problem. Could you please test it?

proguest1942:
Thanks. I used this patch to verify the connection with QBrailleXL several times. It seems works.

DaveMielke:
fixed by 274f8462a7112096f399a903b6059d0d845c2b7b

PR #392, opened 2022-09-22
Stack overflow when connect with bm driver with 64-bits CPU architecture

proguest1942:
There is a crash happens in create KeyBinding array: "Cause: stack pointer is in a non-existent map; likely due to stack overflow." while trying to connect with Orbit reader 20 in 64 bit CPU architecture which the size variable in this case is more than 15,000. It seems static creation is not enough for this case so we propose malloc solution and the stack overflow not exists anymore.

### DIFF
--- a/Programs/ktb_compile.c
+++ b/Programs/ktb_compile.c
@@ -1517,7 +1517,7 @@ addIncompleteBindings (KeyContext *ctx) {
   if (!count) return 1;
 
   size_t size = count * sizeof(*ctx->keyBindings.table);
-  KeyBinding bindings[size];
+  KeyBinding* bindings = (KeyBinding *) malloc(size * sizeof(KeyBinding)) ;
   memcpy(bindings, ctx->keyBindings.table, size);
 
   const KeyBinding *binding = bindings;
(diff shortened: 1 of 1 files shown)


DaveMielke:
Looking at that code, it seems that there's actually a bug. The local copy of the key bindings table should actually be allocated with count - not with size. Please, instead, test this very simple patch.

diff --git a/Programs/ktb_compile.c b/Programs/ktb_compile.c
index 4dca96f3c..10299e072 100644
--- a/Programs/ktb_compile.c
+++ b/Programs/ktb_compile.c
@@ -1517,7 +1517,7 @@ addIncompleteBindings (KeyContext *ctx) {
   if (!count) return 1;
 
   size_t size = count * sizeof(*ctx->keyBindings.table);
-  KeyBinding bindings[size];
+  KeyBinding bindings[count];
   memcpy(bindings, ctx->keyBindings.table, size);
 
   const KeyBinding *binding = bindings;

proguest1942:
Cool, I didn't notice this.
Yes, I patched you solution and the amount of count is about 200. No stack overflow happens again.

DaveMielke:
Thanks for confirmig. And, even more, thanks for figuring out and reporting the problem!

DaveMielke:
fixed by 5ade256558e6c2a532d1bd97d90ba8873c0b8c4a

PR #400, opened 2023-03-30
Fix crash on GetModelIdentifier while driver unconstructed

sthibaul:
If the braille driver is not constructed, disp is NULL, and thus a BrlAPI GetModelIdentifier request would crash. Return an empty string instead in that case.


sthibaul:
For information, with a group of students we are fuzzing the BrlAPI protocol, we will probably find a few of such bugs. We will submit the fuzzing infrastructure at the end of the project (a couple months from now).

shindere:
That's awesome. Many thanks Samuel, and thanks to your students, too.
Would be happy to support in any way (prooof read report, come to the
rehersal of the presnetaiton if there is one...).

PR #405, opened 2023-04-28
proc/*/stat parsing bug

satylogin:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1024811

($comm) can contain space or even ) in its name. So in order to parse it correctly, we should treat comm to be between 1st '(' and last ')'. 

Since this code doesn't really consume $comm, we can just skip it by getting the first ')' from end, and start processing after that.

### DIFF
--- a/Programs/brlapi_client.c
+++ b/Programs/brlapi_client.c
@@ -1662,9 +1662,28 @@ static int getControllingTty(void)
     FILE *stream;
 
     if ((stream = fopen("/proc/self/stat", "r"))) {
-      int vt;
-      int ok = fscanf(stream, "%*d %*s %*c %*d %*d %*d %d", &tty) == 1;
+      char buf[256];
+      size_t read = fscanf(buf, sizeof(*buf), sizeof(buf) / sizeof(buf[0]), stream);
       fclose(stream);
+
+      /* The line format is "$pid ($comm) $state ...
+       * where comm may contain space as part of its name, ex:
+       * 12345 (someproc) S 123 456) R ...
+       * So to process valid identified after comm, we should skip comm by
+       * getting the 1st ) from end.
+       */
+      char *ptr = buf + read - 1;
+      while (*ptr && *ptr != ')' && buf < ptr) {
+        ptr -= 1;
+      }
+      if (*ptr != ')') {
+        break;
+      }
+      /* Point to space after ($comm). */
+      ptr += 1;
+
+      int vt;
+      int ok = sscanf(ptr, " %*c %*d %*d %*d %d", &tty) == 1;
 
       if (ok && (major(tty) == TTY_MAJOR)) {
         vt = minor(tty);
--- a/Programs/brlapi_client.c
+++ b/Programs/brlapi_client.c
@@ -1659,41 +1659,46 @@ static int getControllingTty(void)
 #endif /* WINDOWS */
 
 #ifdef linux
-  {
-    int fd;
-    if ((fd = open("/proc/self/stat", O_RDONLY, 0)) >= 0) {
-      char buf[256];
-      size_t read_bytes = read(fd, buf, sizeof(buf) - 1);
-      close(fd);
-      /* Append null at the end of buffer to ensure that read doesn't
-       * overflows.*/
-      buf[255] = '\0';
-
-      /* The line format is "$pid ($comm) $state ...
-       * where comm may contain space and `)` as part of its name, ex:
-       * 12345 (someproc) S 123 456) R ...
-       * So to process valid identified after comm, we should skip comm by
-       * getting the 1st ) from end.
-       */
-      char *ptr = buf + read_bytes - 1;
-      while (buf < ptr && *ptr != ')') {
-        ptr -= 1;
-      }
-      if (*ptr != ')') {
-        return -1;
-      }
-      /* Point to space after ($comm). */
-      ptr += 1;
-
-      int vt;
-      int ok = sscanf(ptr, " %*c %*d %*d %*d %d", &tty) == 1;
-
-      if (ok && (major(tty) == TTY_MAJOR)) {
-        vt = minor(tty);
-        if ((vt >= 1) && (vt <= MAXIMUM_VIRTUAL_CONSOLE)) return vt;
-      }
+  do {
+    int fd = open("/proc/self/stat", O_RDONLY, 0);
+    if (fd < 0) {
+      break;
     }
-  }
+
+    char buf[256];
+    ssize_t read_bytes = read(fd, buf, sizeof(buf) - 1);
+    close(fd);
+    if (read_bytes < 0) {
+      break;
+    }
+    /* Append null at the end of buffer to ensure that read doesn't
+     * overflows.*/
+    buf[read_bytes] = '\0';
+
+    /* The line format is "$pid ($comm) $state ...
+     * where comm may contain space and `)` as part of its name, ex:
+     * 12345 (some proc (foo) S 123 456) R ...
+     * So to process valid identified after comm, we should skip comm by
+     * getting the 1st `)` from end.
+     */
+    char *ptr = buf + read_bytes - 1;
+    while (ptr > buf && *ptr != ')') {
+      ptr -= 1;
+    }
+    if (*ptr != ')') {
+      break;
+    }
+    /* Point to space after ($comm). */
+    ptr += 1;
+
+    int vt;
+    int ok = sscanf(ptr, " %*c %*d %*d %*d %d", &tty) == 1;
+
+    if (ok && (major(tty) == TTY_MAJOR)) {
+      vt = minor(tty);
+      if ((vt >= 1) && (vt <= MAXIMUM_VIRTUAL_CONSOLE)) return vt;
+    }
+  } while (0);
 #endif /* linux */
 
   return -1;
--- a/Programs/brlapi_client.c
+++ b/Programs/brlapi_client.c
@@ -36,6 +36,7 @@
 #include <locale.h>
 #include <assert.h>
 #include <limits.h>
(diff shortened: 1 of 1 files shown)


satylogin:
What would be the best way to test these changes locally ?

Programs/brlapi_client.c
```diff
@@ -1662,10 +1662,29 @@ static int getControllingTty(void)
...
-      int ok = fscanf(stream, "%*d %*s %*c %*d %*d %*d %d", &tty) == 1;
+      char buf[256];
+      size_t read = fscanf(buf, sizeof(*buf), sizeof(buf) / sizeof(buf[0]), stream);
       fclose(stream);
 
+      /* The line format is "$pid ($comm) $state ...
+       * where comm may contain space as part of its name, ex:
+       * 12345 (someproc) S 123 456) R ...
+       * So to process valid identified after comm, we should skip comm by
+       * getting the 1st ) from end.
```
sthibaul, line 1673: I'd rather parse for ')' starting from the beginning, since we don't know what could be introduced at the end of this file

satylogin: `)` could also be a valid character of comm (example in comment). If this was parsed with 1st ')' from begining, that would lead to state being S and pid being 123 and so on... which would also be incorrect.

The reason for choosing ')' from end is the fact that whatever follows comm in an integer value or whitespace (https://man7.org/linux/man-pages/man5/proc.5.html, from: `/proc/[pid]/stat`).

Programs/brlapi_client.c
```diff
@@ -1662,10 +1662,29 @@ static int getControllingTty(void)
     FILE *stream;
 
     if ((stream = fopen("/proc/self/stat", "r"))) {
-      int vt;
-      int ok = fscanf(stream, "%*d %*s %*c %*d %*d %*d %d", &tty) == 1;
+      char buf[256];
+      size_t read = fscanf(buf, sizeof(*buf), sizeof(buf) / sizeof(buf[0]), stream);
```
sthibaul, line 1666: you meant fread here?

sthibaul: Since we'd just read from the file and not scan, better just replace with open/read/close

sthibaul: (also, with nowadays' C standard, sizeof(char) is always 1)

sthibaul: > you meant fread here?

(put another way: please *test* your changes...)

sthibaul: (as in: at the very least check that building doesn't spit warnings)

sthibaul: As for actual testing, you can make the function non-static, and call it as a library function, it will work independently of the rest of the library.

Programs/brlapi_client.c
sthibaul: > `)` could also be a valid character of comm (example in comment). 

Uh... (now I have read the debian bug...)

> whatever follows comm is an integer value or whitespace

That's what it is now only, things can and have been added there. I see people saying on the Internet that adding a string with `)` would make it ambiguous and thus that can't happen. Well, the kernel was supposed to respect the documentation (can be parsed with %s) but they didn't.

I guess we indeed have no choice but parse `)` from the right indeed.

sthibaul: Then also update the comment: not only `where comm may contain space as part of its name`, but also mention right parentheses.

Programs/brlapi_client.c
```diff
@@ -1662,10 +1662,29 @@ static int getControllingTty(void)
...
       fclose(stream);
 
+      /* The line format is "$pid ($comm) $state ...
+       * where comm may contain space as part of its name, ex:
+       * 12345 (someproc) S 123 456) R ...
+       * So to process valid identified after comm, we should skip comm by
+       * getting the 1st ) from end.
+       */
+      char *ptr = buf + read - 1;
+      while (*ptr && *ptr != ')' && buf < ptr) {
```
sthibaul, line 1676: The `*ptr` test is meaningless: `\0` appear at end of strings, not at beginning of strings.

Programs/brlapi_client.c
```diff
@@ -1662,10 +1662,29 @@ static int getControllingTty(void)
...
+        ptr -= 1;
+      }
+      if (*ptr != ')') {
+        break;
```
sthibaul, line 1680: you meant `return -1`?

Programs/brlapi_client.c
sthibaul, line 1666: You need to add a trailing `\0` at the end of the read string (and thus substract 1 from the size passed to read), otherwise `sscanf` below may overflow.

Programs/brlapi_client.c
satylogin: Totally agree !!

Programs/brlapi_client.c
satylogin: Ahh !! Yeh... Thanks !!

Programs/brlapi_client.c
satylogin: Updated.

Programs/brlapi_client.c
satylogin: Updated.

satylogin: Tested the function body separately.

satylogin: Do we also have some github actions or sandbox tests that would ensure correctness ?

Programs/brlapi_client.c
```diff
@@ -1659,12 +1660,33 @@ static int getControllingTty(void)
 
 #ifdef linux
   {
-    FILE *stream;
+    int fd;
+    if ((fd = open("/proc/self/stat", O_RDONLY, 0)) >= 0) {
+      char buf[256];
+      size_t read_bytes = read(fd, buf, sizeof(buf) - 1);
```
sthibaul, line 1666: `read` returns an `ssize_t`, better use that, and check that it suceeded.

To simplify the whole thing, maybe use a `do { } while(0)` block around the whole thing, so you can just `break` in error cases

Programs/brlapi_client.c
```diff
@@ -1659,12 +1660,33 @@ static int getControllingTty(void)
...
+      buf[255] = '\0';
+
+      /* The line format is "$pid ($comm) $state ...
+       * where comm may contain space and `)` as part of its name, ex:
+       * 12345 (someproc) S 123 456) R ...
+       * So to process valid identified after comm, we should skip comm by
+       * getting the 1st ) from end.
+       */
+      char *ptr = buf + read_bytes - 1;
+      while (buf < ptr && *ptr != ')') {
```
sthibaul, line 1679: I'd say rather use `ptr >= buf`, that'll look more idiomatic

sthibaul (changes requested):

Programs/brlapi_client.c
```diff
@@ -1659,12 +1660,33 @@ static int getControllingTty(void)
...
+      char buf[256];
+      size_t read_bytes = read(fd, buf, sizeof(buf) - 1);
+      close(fd);
+      /* Append null at the end of buffer to ensure that read doesn't
+       * overflows.*/
+      buf[255] = '\0';
+
+      /* The line format is "$pid ($comm) $state ...
+       * where comm may contain space and `)` as part of its name, ex:
+       * 12345 (someproc) S 123 456) R ...
```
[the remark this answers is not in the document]
satylogin: Updated

Programs/brlapi_client.c
satylogin: Update to `ptr > buf`, since if we do `>=` check and for some reason the file is not formatted correctly, we may end up at buf - 1, and the next if checks for *(buf - 1), which is something that we should probably avoid. If we are at `buf + 1` and we still haven't found ')', then even if the prev char is ')' that would still be invalid format, so it should be safe to avoid.

Let me know if this sounds right.

Programs/brlapi_client.c
satylogin: Done

Programs/brlapi_client.c
```diff
@@ -1658,20 +1659,46 @@ static int getControllingTty(void)
...
-      int ok = fscanf(stream, "%*d %*s %*c %*d %*d %*d %d", &tty) == 1;
-      fclose(stream);
+    char buf[256];
+    ssize_t read_bytes = read(fd, buf, sizeof(buf) - 1);
+    close(fd);
+    if (read_bytes < 0) {
+      break;
+    }
+    /* Append null at the end of buffer to ensure that read doesn't
+     * overflows.*/
```
sthibaul, line 1675: It's not read that would overflow, but sscanf

Programs/brlapi_client.c
```diff
@@ -1658,20 +1659,46 @@ static int getControllingTty(void)
...
+    buf[read_bytes] = '\0';
+
+    /* The line format is "$pid ($comm) $state ...
+     * where comm may contain space and `)` as part of its name, ex:
+     * 12345 (some proc (foo) S 123 456) R ...
```
sthibaul, line 1680: you can drop ` S 123 456`, the ` (foo)` part alone is enough

sthibaul (changes requested):

Programs/brlapi_client.c
satylogin: Done

Programs/brlapi_client.c
satylogin: Done

sthibaul:
@DaveMielke this looks good to me

sthibaul:
@satylogin Thanks!

satylogin:
Thanks for your help with the reviews !!

PR #417, opened 2023-05-12
Help Tech driver updates

FelixGruetzmacher:
The duplicate key bindings in the Activator key table are now fixed. There was a conflict with an include.


DaveMielke:
merged via commit c5ac6e016c23bc488de78e086d15ab0f84270a14

PR #422, opened 2023-07-10
ensure appending to clipboard with a rectangular copy inserts a CR

npitre:
Concatenating rectangular copy should be done vertically i.e. a CR
should be present to separate it from the existing clipboard content.

### DIFF
--- a/Programs/cmd_clipboard.c
+++ b/Programs/cmd_clipboard.c
@@ -89,9 +89,25 @@ cpbReadScreen (ClipboardCommandData *ccd, size_t *length, int fromColumn, int fr
 }
 
 static int
-cpbEndOperation (ClipboardCommandData *ccd, const wchar_t *characters, size_t length) {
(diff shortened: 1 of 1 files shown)


shindere:
npitre (2023/07/10 13:55 -0700):
> Concatenating rectangular copy should be done vertically i.e. a CR
> should be present to separate it from the existing clipboard content.

Isn't that use-case oriented?

Why favour e.g. enter over space?

npitre:
Linear copy is "linear" i.e. multiple lines are collapsed to a single 
line with spaces.

Rectangular copy is a block. They should be appended vertically.

On the command line, you can easily convert the later to the former with 
"echo" and/or "xargs".

shindere:
npitre (2023/07/10 14:12 -0700):
> Linear copy is "linear" i.e. multiple lines are collapsed to a single 
> line with spaces.

Is this what you have implemented? Because at the moment, I don't thik a
space is added during linear copies, or am I mistaken here?

PR #423, opened 2023-07-19
systemd: add user brltty only once to group tty

olafhering:


DaveMielke:
fixed properly by commit 947ebc99246a59995cc753487b3749567000d358

PR #437, opened 2023-10-20
Support building Windows version on Appveyor

leonardder:
As I have been struggling for ages to get a working windows build of BRLTTY for myself, I figured building the Windows version with Continuous Integration would be most helpful.
Appveyor has baked in support for msys1, so that seemed like a good starting point, and indeed, it works like a charm.
The bundled configuration for Appveyor builds BRLTTY and uses Python 3.11 for the Python bindings. The resulting zip and installer executable are uploaded as artifacts and can be fetched from the build after building.

i have also added Unicode false to the nsis script. This is the default behavior in the version of nsis that is used by the build system by default, but Appveyor has a newer version that has Unicode set to true by default. When we change Unicode to true, we get errors with nsistrings.txt parsing.

Note that you need to setup Appveyor for this to work, which is pretty trivial:

1. Go to https://www.appveyor.com/
2. Sign in with Github
3. Create a new project from Github
4. Login with your Github credentials
5. Install the Appveyor app for the BRLTTY organization for the brltty repo.

When setup is finished, pushing a next commit to master or triggering the build from the appveyor dashboard should work. It should also work for pull requests out of the box.

note that this can be expanded in many ways, i.e:

1. Building for both versions of LibUSB
2. Building for several versions of Python
3. Building for Linux

### DIFF
--- a/Windows/brltty.nsi
+++ b/Windows/brltty.nsi
@@ -35,6 +35,7 @@
 		!define DISTDIR "brltty-win-${VERSION}-${VARIANT}"
 	!endif
 
+	Unicode false
 	SetCompressor /SOLID LZMA
 	SetOverwrite IfNewer
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,54 @@
+###############################################################################
+# BRLTTY - A background process providing access to the console screen (when in
+#          text mode) for a blind person using a refreshable braille display.
+#
+# Copyright (C) 1995-2023 by The BRLTTY Developers.
+#
+# BRLTTY comes with ABSOLUTELY NO WARRANTY.
+#
+# This is free software, placed under the terms of the
+# GNU Lesser General Public License, as published by the Free Software
+# Foundation; either version 2.1 of the License, or (at your option) any
+# later version. Please see the file LICENSE-LGPL for details.
+#
+# Web Page: http://brltty.app/
+#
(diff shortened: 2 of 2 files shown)


DaveMielke:
Should the .yml file be at the top-level or in Windows/?

What's a simple description of what Appveyor does?

-- 
EMail: ***@***.***  | Ottawa, Ontario   | Twitter: @Dave_Mielke

leonardder:
> Should the .yml file be at the top-level or in Windows/? 

By default, it should be in the top level directory. You can also move it to the Windows directory, but that requires changing the configuration in Appveyors gui, i.e. opening project settings and on the general tab, appveyor.yml Custom configuration .yml file name, enter the new path. However, if you ever wish to add appveyor for linux builds as well, you'd have to undo this.

> What's a simple description of what Appveyor does?

To keep it very simple, when Appveyor is setup and you commit something to the master branch, Appveyor spins up a Windows virtual machine and runs the instructions from appveyor.yml to build your code. In the end, when the build finishes, it can publish build artifacts, such as in the case of BRLTTY, the zip and exe files resulting from the Windows build.

See [here](https://ci.appveyor.com/project/leonardder/brltty/builds/48331160) for an example run. You can retrieve the build artifacts when going to the [Artifacts](https://ci.appveyor.com/project/leonardder/brltty/builds/48331160/artifacts) link.

PR #476, opened 2025-05-05
Add the address parameter to the speech-dispatcher speech driver

shindere:
This subsumes the removed port parameter.

If the address parameter is passed to the driver, its contents gets
put in the SPEECHD_ADDRESS environment vriable so that libspeechd
can use it. If the environment variable has already been set, the
driver parameter does not overwrite it.

This implements a suggestion from @sthibaul


shindere:
Dave Mielke (2025/05/05 09:25 -0700):
> Merged #476 into master.

Thanks!

PR #491, opened 2025-08-31
dracut: move the module ordering from 99 to 50

jozzsi:
Follow the recommended dracut module ordering for out-of-tree modules.

For dracut release v108 or later the recommended ordering for out out of tree modules is 50-59 range. The following is a section from dracut documentation:

> Not using the 50-59 range for out of tree dracut modules will likely
> lead to unintended errors in the initramfs generation process as your
> dracut module will either run too early or too late in the generation process.
> You have been warned.


jozzsi:
I am not able to support feedbacks on this PR any more, so closing to manage expectations. I hope someone else can pick it up (and re-upload or re-open).

PR #494, opened 2025-10-04
Start the graphical BRLTTY like the root one

shindere:
The instance of BRLTTY that follows graphical screens was run with
both speech support and BrlAPI server disabled. This commit removes these
hard-coded assumptions from the startup script.

@sthibaul any opinion on the change about `-N`?

Actually I see no reason why the BrlAPI server should be disabled for the X
brltty?

For the speech driver, disabling it for graphical BRLTTY definitely feels
wrong, as if a user
uses speech in the text consoles it is very likley that they will want to
have it in graphical terminals,
too.

### DIFF
--- a/Autostart/X11/90xbrlapi.in
+++ b/Autostart/X11/90xbrlapi.in
@@ -6,6 +6,7 @@ drivers_directory="@drivers_directory@"
 program_directory="@program_directory@"
 xbrlapi="$program_directory/xbrlapi"
 brltty="$program_directory/brltty"
+xbrltty_flags="-b ba -x a2 -N"
 
 if [ -x "${xbrlapi}" ]; then
   if "${xbrlapi}" 2>/dev/null ; then
@@ -13,7 +14,7 @@ if [ -x "${xbrlapi}" ]; then
     if [ -x "${brltty}" -a \
          -e "$drivers_directory/libbrlttybba.so" -a \
          -e "$drivers_directory/libbrlttyxa2.so" ]; then
-      "${brltty}" -b ba -s no -x a2 -N 2>/dev/null
+      "${brltty}" ${xbrltty_flags} 2>/dev/null
     fi
   fi
 fi


sthibaul:
> I see no reason why the BrlAPI server should be disabled for the X brltty?

The textual brltty already serves brlapi, and the graphical one uses it. The graphical brltty will only fail to set up another one on the same socket. And setting it up on another socket will most probably just bring confusion to users.

shindere:
Okay thanks, it was the middle of the night and it's probably why I
thought there could be cases where one would want xbrltty to act as some
form of proxy.

I have updated both the code and the commit message. Thanks.

shindere:
My understanding is that if you ouse BRLTTY to render a terminal in
braille, then it is very likely that, if you want speech rendering at
all, you will want this speech rendering to be done by BRLTTY rather
than Orca just because it will mean that the speech rendering is
synchronized with the braille display.

Also, the way things are currently is odd because the current behaviour
means that, even if you enable speech in the root brltty that follows
the virtual consoles, you own't have speech when you are in a graphical
terminal and my experience is that it has been rather difficult to
understand why this is so.

I agree that the situation that will result from this change may look
strange from  the outside, but I would say that the current situation is
already strange and that the impact on users will be small as few of
them do actually use this setup at the moment.

If unsure we can discuss the matter on the list.

sthibaul:
I had added that `-s no` only to be the least disruptive as possible, by just enabling braille, whose prioritization is well under control, while for speech I didn't know how that would work, so I had preferred to leave it disabled to avoid disruption.

shindere:
Dave Mielke (2025/10/05 01:35 -0700):
> DaveMielke left a comment (brltty/brltty#494)
> 
> Also, we must remember that telling Orca not to speak means that it
> won't even speak for non-terminal windows.

No, Orca can have per-applicaiton settings. See the messages I sent on
the list where I explain this.

There is indeed still kind of a problem, I think, whihc is the
accessibility of the other widgets of the application providing the
terminal.

DaveMielke:
Per application is even worse. Turn off speech for gnome-terminal, for mate-terminal, for xterm, etc, etc, etc. Very tedious. The default configuration must never be tedious, or, really, even mildly inconvenient - it should just work. If the default for Orca is to speak, and if we don't want double speech, then the default for brltty must be to not speak. Of course, there'd be nothing wrong with having a way to tell brltty not to speak but, at least to me, not speaking must remain the default.

-- 
EMail: ***@***.***  | Ottawa, Ontario   | Twitter: @Dave_Mielke
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
http://mielke.cc/xmother.html (Letter from a Feminist ex-Mother)

shindere:
The experience of braille rendering being done by one screen reader and
speech rendering being done ay another is not going to be great.

Of course, whatever default gets chosen, it remains possible for users
to override it by means of configuration files, but the way things are
done currently makes it so that the two instances of brltty share the
configuration file.

May I however suggest that we continue the discussion on the lists so
that everybody can (more easily) participate?

sthibaul:
> Turn off speech for gnome-terminal, for mate-terminal, for xterm, etc, etc, etc.

There are really only gnome-terminal and mate-terminal which are accessible. xterm will essentially never be, and others will probably not care either. Possibly konsole seems to be exposing some text but it seems to be completely bogus, and thus unusable.

One user will use only one kind of terminal anyway, so they won't have to care about disabling speech for all terminals

DaveMielke:
[quoted lines by Sébastien Hinderer on 2025/10/05 at 02:30 -0700]

>The experience of braille rendering being done by one screen reader and
>speech rendering being done ay another is not going to be great.

I understand. I'm not saying that we shouldn't allow it - I'm saying that configuration should default to something that just works. If we could control Orca then that'd be a different matter.

>Of course, whatever default gets chosen, it remains possible for users
>to override it by means of configuration files, but the way things are
>done currently makes it so that the two instances of brltty share the
>configuration file.

Yes. I have no problem figuring out a way to do it with two.

>May I however suggest that we continue the discussion on the lists so
>that everybody can (more easily) participate?

Sure. I just respond to messages where I get them. If I'm sent a message from another place then I'll respond to it there. One thing I'm a bit sensitive to, though, is that I don't want anyone to think that the final word has been given just because I state my opinion.

-- 
EMail: ***@***.***  | Ottawa, Ontario   | Twitter: @Dave_Mielke
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
http://mielke.cc/xmother.html (Letter from a Feminist ex-Mother)

PR #495, opened 2025-10-05
Fix crash on speech dispatcher connexion error

sthibaul:
spd_open2 needs the address of a pointer that it will fill. Passing NULL is deemed to fail.


sthibaul:
Cc @shindere

PR #532, opened 2026-05-17
Darwin: register FileViewer / Tmux / Screen / TerminalEmulator drivers

yplassiard:
Two small additions on top of #531 that round out the macOS support.

### brltty-pty: skip the TTY check when driving the `em` screen driver

brltty-pty's `main()` refuses to run if `stdin` / `stdout` aren't TTYs — the right default for the standalone `brltty-pty bash` invocation. But the `em` screen driver also spawns brltty-pty as a child (passing `--driver-directives`) to obtain the SHM-backed terminal segment, and in that mode the parent (brltty under sudo) has no controlling TTY: stdin/stdout are pipes and the isatty checks abort brltty-pty before publishing its path. The em driver then sees the emulator stream close immediately and shuts brltty down.

Skip the TTY checks when `--driver-directives` is set; that flag already signals daemon-spawned mode, where input arrives via the message queue rather than stdin. Standalone invocations are unaffected.

### configure.ac: enable FileViewer / Tmux / Screen / TerminalEmulator on darwin

In addition to MacOSAccessibility (which stays the default screen driver), build the four console-style screen drivers on macOS:
- `fv` (FileViewer) is what brltty uses to render its built-in help and learn-mode pages. Setupterm/tigetstr/... live in `libncurses` on macOS (no separate `libtinfo`), so the darwin branch links against `-lncurses`.
- `em` (TerminalEmulator) paired with brltty-pty bypasses the AX pipeline entirely — useful both as a fast-path for terminal-only workflows and as a benchmarking target.
- `sc` (Screen) and `tx` (Tmux) build cleanly once `em` is enabled.

`DEFAULT_SCREEN_DRIVER` on macOS stays `mo` (the first one registered under the darwin case).

### Stack

Builds on #529, #530, #531.


DaveMielke:
I don't understand this comment in configure.ac:

   dnl FileViewer is required by brltty's built-in help / learn-mode UI

I don't believe this to be true. Please explain.

Also, configure.ac is bad enough without duplicated code. Is the only difference between those four screen drivers that MacOS doesn't want -ltinfo?

-- 
I believe the Bible to be the very Word of God. -> http://Mielke.cc/bible/
Dave Mielke            | 2213 Fox Crescent | Amator et provisor concubitus uxori meae
EMail: ***@***.***  | Ottawa, Ontario   | Home Page: http://Mielke.cc/

The most important document I've ever written. -> http://mielke.cc/ruler.html
concubitus est ut maritus uxori suae beneplacitum, satisfactionem, et contentum praebeat
cum uxor suaviter tractatur, cunnus eius multo maiorem remissionem mentulae mariti praebebit

Issue #540, opened 2026-06-21
First-class braille access in GUI terminals (and macOS support)


hlub:
This is a tracking issue for a set of changes that make BRLTTY usable from
**graphical terminal emulators** — and, as the headline use case, make BRLTTY
work on **macOS**, where there is no Linux-style text console at all.

The enabling component is BRLTTY's existing **Terminal Emulator screen driver**
(`scr.em`) and its companion **`brltty-pty`**: `brltty-pty` runs a shell (or any
program) inside a pseudo-terminal, emulates the terminal, and publishes the
screen to BRLTTY via shared memory, so the braille display can read it. That is
the only way to get a braille-accessible "screen" inside a GUI terminal (e.g.
Terminal.app, iTerm2, GNOME Terminal, Konsole), and the only practical way to use
BRLTTY on macOS.

Today that path is rough: keys like Alt/Home/arrows misbehave, the screen is
stuck at 80×24, resizing doesn't work, some Unicode/box-drawing is wrong, and
common programs (tmux, vim, Claude Code, anything that probes the terminal)
misbehave or hang. The PRs below fix that, plus the macOS-specific plumbing
(async I/O, Bluetooth input, build) needed to run BRLTTY there at all.

## What users get

- **Blind macOS users** can run BRLTTY with a refreshable braille display and use
  a real shell, `vim`, `tmux`/`screen`, `less`, and modern TUIs (e.g. Claude
  Code) inside Terminal.app / iTerm2 — keyboard, resizing, Unicode, and clipboard
  all working.
- **GUI-terminal users on every platform** (not just macOS): the `brltty-pty`
  improvements are platform-neutral, so anyone who prefers a braille-accessible
  terminal *inside* their graphical session — rather than a Linux VT — gets a far
  more correct experience.
- **The on-screen image is finally correct too.** brltty-pty now lets the
  terminal render the program's output directly (PR #546), so colours and styling
  appear exactly as intended — full true-colour, not the 16-colour approximation
  curses re-rendering produced. This matters whenever a sighted person works
  alongside the braille user (teaching, pairing, support).

## Why the terminal-emulator approach is the right one

A screen reader inside a GUI terminal only needs to *observe* one program's
screen to drive braille — it is not a multiplexer. Two design choices follow, and
they are the spine of this work:

1. **Read the screen from a shared-memory grid, not by scraping pixels.**
   `brltty-pty` interprets the child's output into a cell grid and publishes it to
[...]

PR #544, opened 2026-06-21
brltty-pty: robust terminal emulation (input, resize, UTF-8, OSC 52, tests)

hlub:
Closes part of #540.

This is the flagship change: a rewrite of the brltty-pty terminal-emulation path that makes it correct for daily use.

**What was broken:**
- Special keys (Alt, Home, arrows, function keys) were not forwarded to the child
- Terminal resize events were not propagated
- Unicode and box-drawing characters were mishandled
- Programs that probe the terminal (tmux, vim, Claude Code, ncurses apps) misbehaved or hung

**What this PR adds:**
- Correct input forwarding: raw key sequences pass through to the child via the PTY
- Terminal resize: SIGWINCH propagated; brltty-pty tracks and reports the current size
- UTF-8 stream decoder: multi-byte characters are assembled correctly before being written to the screen grid
- OSC 52: programs running inside brltty-pty can copy to the host clipboard via the standard OSC 52 escape (file-hook fallback for terminals that do not support it)
- macOS termination fix: folded in (the process-group signal that was being lost)
- First brltty-pty(1) man page
- End-to-end test programs (ptytest, ptyinput, ptyresize) built by make tests

**Platform scope:** fully cross-platform. The macOS termination fix is the only Apple-specific piece (one ifdef).

**Testing:** make tests passes on macOS 15 (arm64) and Linux (x86_64). Manually tested with vim, tmux, less, and Claude Code inside brltty-pty.


hlub:
Added 4 more commits after further real-world use and a code review turned up two related freeze bugs, both reproducible:

- **brltty-pty could wedge the whole terminal emulator indefinitely.** Braille-injected keystrokes were written to the child with a blocking `write()` on brltty-pty's single event-loop thread — the same thread that reads the child's output. If the child wasn't draining input right when this happened (e.g. mid-repaint), the write could block forever, freezing rendering with no way to recover. Fixed by routing the write through the existing `asyncWriteFile()` queuing mechanism.
- **A leaked SysV shared-memory segment/message queue could hang unrelated processes system-wide.** brltty-pty only removes its IPC resources on a clean exit; an uncleanly-killed instance (crash, SIGKILL) leaks both. SysV message-queue limits are tiny system-wide, so a handful of leaks exhaust the whole budget — after which blocking IPC calls in *any* process, not just brltty-pty, hang until reboot. Reproduced live: one leaked queue caused a fresh, unrelated process's `msgsnd()` to hang indefinitely. Fixed with a small file-based registry: each instance reaps IPC objects left behind by dead PIDs at startup.
- A third commit hardens the screen driver's polling (a dropped push notification no longer sticks the display on stale content). A fourth, review-driven commit closes a privilege-boundary gap found in the reaper (verifies SysV object ownership before removing it, so a local user can't trick a root-run brltty-pty into destroying an unrelated object) and restores a write-failure diagnostic the async-write fix had dropped.

**Testing:** `make tests` passes (macOS arm64). The leak and its fix were verified live: manually registered a leaked shm/msg pair under a dead PID and confirmed the next brltty-pty startup reaps it.

hlub:
Added 3 more commits after tracking down a related freeze.

One of them (the segment-updated notification) turned out to already be
fixed on macos/integration but never actually made it into this PR - sorry
about that, catching it up now:

- **The segment-updated notification (TERM_MSG_SEGMENT_UPDATED) was still
  being sent as a blocking msgsnd.** Switched it to IPC_NOWAIT: the driver
  also polls the segment, so a dropped "updated" hint just defers a braille
  refresh, and blocking here could stall the whole I/O loop (and the child)
  whenever the daemon is slow and the small SysV queue fills during an
  output burst. Folded in alongside this: logging (instead of silently
  dropping) an OSC string that overflows the 64 KiB OSC buffer, removing a
  write-only dead variable, and a couple of typo fixes.
- **brltty-pty's stderr (a pipe back to BRLTTY when it isn't a real
  terminal) could block the whole event loop.** If BRLTTY's own main thread
  ever stalls for any reason, it stops draining that pipe; once full,
  brltty-pty's next log write blocks - on the same single thread that
  drains the child's pty and renders its output, freezing the child along
  with it. The actual fuel filling that pipe under sustained backpressure
  was the previous bullet's IPC_NOWAIT path logging once per dropped
  message. Fixed by making the stderr pipe non-blocking (a full pipe now
  drops a log line instead of blocking the process) and rate-limiting that
  logging to once on entering/leaving a failure streak.
- **The pty master was being read one byte at a time.** Bumped the read
  buffer from 1 byte to 4096; read() still returns as soon as anything is
  available, so interactive latency is unaffected, but a burst of child
  output is now processed in one pass instead of thousands of tiny ones -
  which also directly reduces how often the message-queue path above can
  even be triggered. Fixed a latent bug this exposed: the per-batch
  refresh-wanted flag was being overwritten, not OR'd, across a batch, so an
  earlier byte's refresh request could be lost if a later byte in the same
  batch didn't also want one - harmless at the old 1-byte batch size, would
  have silently dropped screen refreshes at the new one.

Testing: this exact logic (byte-for-byte identical) has been running on
macos/integration, where I reproduced the original freeze live - a process
running inside brltty-pty stopped making any progress at all while the
braille display was disconnected - and confirmed all three commits fix it
(the process keeps running normally through the disconnect, and
reconnecting the display afterward recovers cleanly), plus a clean
`make tests` run there. I could not do a from-scratch build of this branch
in isolation this round (missing some unrelated macOS-only autostart
scaffolding that only exists on my full working branch) so I'm not
separately claiming a `make tests` run against this exact commit set, but
each changed file compiles cleanly on its own against this branch's code.

hlub:
Added two more commits, from real-world testing after the previous batch landed:

- **A screen-attach race**: emEmulatorMonitor() read one line per readable-fd wakeup, but fully-buffered stdio can pull more than one already-written line out of the pipe in a single read - e.g. a routine log line queued right in front of the initial "path" handshake. Once that happens the second line is invisible to select()/poll() and gets stranded indefinitely. Now drains every buffered line per wakeup instead of just one.
- **A shared-memory leak survivor**: the startup IPC reaper only ever knew about a segment's *original* identifier; ptyResizeScreen() publishes a new identifier on every resize but never re-registered it, so a resize followed by an unclean exit leaked the segment permanently instead of being reapable. Confirmed live: nine orphaned segments accumulated in one evening of testing. Also added SIGHUP handling, since closing the terminal window (by far the most common way this process ends) was hitting exactly this gap every time.

PR #547, opened 2026-06-21
brltty-handoff: BrlAPI utility to suspend/resume the braille driver + macOS automation docs

hlub:
Closes part of #540.

This PR adds a new command-line utility and the macOS integration docs/automation that depend on it.

**brltty-handoff** (5 commits):
- A small BrlAPI utility: brltty-handoff suspend tells BRLTTY to release the braille display (so another application, e.g. VoiceOver on macOS, can claim it), and brltty-handoff resume re-claims it. Operates via the BrlAPI driver-directive interface, no root required.

**README.macOS** — a new document describing how to set up BRLTTY on macOS: prerequisites, build, install, brltty-pty usage, Bluetooth pairing, clipboard configuration.

**Hammerspoon automation** — an optional Lua script for Hammerspoon (macOS automation app) that automatically hands off the braille display between BRLTTY and VoiceOver when focus moves between terminal and other apps:
- Detects which terminal started brltty and watches for app-switch events
- Calls brltty-handoff to suspend/resume and launches/quits VoiceOver as needed
- Comes with a Makefile install target and a sudoers template so brltty-handoff can run without a password prompt

**Recommend splitting** (as noted in #540): brltty-handoff itself could be reviewed independently of the macOS-specific docs/automation. Happy to split into two PRs if preferred.

### DIFF
--- a/Autostart/Hammerspoon/Makefile.in
+++ b/Autostart/Hammerspoon/Makefile.in
@@ -1,5 +1,3 @@
-#!/bin/sh
-# @configure_input@
 ###############################################################################
 # BRLTTY - A background process providing access to the console screen (when in
 #          text mode) for a blind person using a refreshable braille display.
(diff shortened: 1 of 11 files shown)


hlub:
Added one more commit: `brltty-handoff resume` now waits briefly before calling `brlapi_resumeDriver()`.

Root-caused live via bluetoothd's own logs: a resume typically races the very VoiceOver disconnect it follows, since the daemon takes ~150-200ms to actually finish tearing down the previous session on the same link. Calling resume immediately loses that race almost every time, and - confirmed independently with a minimal reproduction outside BRLTTY - the resulting failure never delivers a completion callback back to the caller at all, so it costs a full timeout-and-retry cycle instead of connecting on the first attempt. A short deliberate wait avoids the race instead of paying for it.

PR #549, opened 2026-06-29
BrlAPI: make the sockets directory 0755 instead of world-writable + sticky

alex19EP:
## Summary

The BrlAPI sockets directory (`BRLAPI_SOCKETPATH`, e.g. `/var/lib/BrlAPI`) was
created **world-writable with the sticky bit set** — `3777` in the systemd
tmpfiles entry, `1777` at runtime. This reduces it to a strict **`0755`**,
owned by the brltty user, removing the world-write and sticky bits (and, in
privileged Linux deployments, the group-write + setgid that the privilege code
would otherwise add).

## Why

That `/tmp`-style mode only existed so a BrlAPI **server** running as a user
*other than* the directory's owner could create its sockets in the shared
directory, with the sticky bit stopping users from clobbering each other's
sockets.

In the normal deployment this is unnecessary and is an avoidable exposure:
BrlAPI is **one system daemon** (which *owns* the directory) serving **many
clients running as different users**. A client only needs to **traverse** the
directory (`x`) and **write to the socket inode** — which the server itself
widens to be connectable. Actual access control is the authorization key
(`/etc/brltty/brlapi.key`, `0640 root:brlapi`), not the directory mode. So the
directory never needs to be writable by anyone but its owner.

## What changed

Four independent mechanisms set this directory's mode; all are tightened so the
result is `0755` everywhere (editing only the tmpfiles entry would have been
silently undone at runtime):

| File | Change |
|------|--------|
| `Programs/brlapi_server.c` | Create the directory `0755` (no sticky); `adjustPermissions()` no longer mirrors owner-write to group/other or sets the sticky bit **for directories**. The socket **file** is still widened so cross-user clients can `connect()`. |
| `Programs/file.c`, `Headers/file.h` | `createDirectory()`'s widening branch now yields `0755` rather than `1777`; the `worldWritable` flag is renamed `worldTraversable`. |
| `Programs/pgmprivs_linux.c` | `claimStateDirectory()` no longer unconditionally adds group-write + setgid; a per-entry `groupWritable` flag exempts the sockets directory while leaving the `updatable` (`2770`) and `writable` (`0750`) directories byte-for-byte unchanged. |
| `Autostart/Systemd/tmpfiles.in` | `3777` → `0755`. |
| `Documents/README.Linux` | Document the new rationale; drop the world-writable/sticky guidance. |

## Behavior change / compatibility

- Cross-user **clients** are unaffected — directory traversal + the
  world-connectable socket inode + key auth all still work at `0755`.
- A BrlAPI **server** run by a user *other than* the directory's owner can no
  longer create its sockets in the shared directory. This degrades gracefully
  (the existing error paths log and bail — no crash/hang). Restoring per-user
  servers with proper isolation (e.g. `$XDG_RUNTIME_DIR/brltty`) is left as
  possible follow-up; this PR is intentionally scoped to hardening the existing
  shared directory and keeps the path at `/var/lib/BrlAPI`.

## AI Usage Disclosure

This change was developed with assistance from Claude (Anthropic). All code
was reviewed and tested by the author before submission.

### DIFF
--- a/Programs/brlapi_client.c
+++ b/Programs/brlapi_client.c
@@ -788,12 +788,12 @@ static int tryHost(brlapi_handle_t *handle, const char *hostAndPort) {
 
 #if defined(PF_LOCAL)
   if (handle->addrfamily == PF_LOCAL) {
-    int lpath = strlen(BRLAPI_SOCKETPATH),lport;
-    lport = strlen(port);
+    int lport = strlen(port);
 
 #ifdef __MINGW32__
     {
       HANDLE pipefd;
+      int lpath = strlen(BRLAPI_SOCKETPATH);
       char path[lpath+lport+1];
 
       memcpy(path, BRLAPI_SOCKETPATH, lpath);
@@ -815,38 +815,86 @@ static int tryHost(brlapi_handle_t *handle, const char *hostAndPort) {
 #else /* __MINGW32__ */
     {
       struct sockaddr_un sa;
+      char xdgPath[sizeof(sa.sun_path)];
+      const char *directories[3];
+      unsigned int directoryCount = 0;
+      unsigned int directoryIndex;
+      int connectError = 0;
 
-      if (lpath+1+lport+1 > sizeof(sa.sun_path)) {
-	brlapi_libcerrno=ENAMETOOLONG;
-	brlapi_errfun="path";
-	brlapi_errno = BRLAPI_ERROR_LIBCERR;
-	goto out;
+      /* Connect to the first server that's found, searching in this order: one
+       * the user started within their own session (under $XDG_RUNTIME_DIR), then
+       * the system server, and finally the legacy location in case a running
+       * server still uses the pre-/run path. So a session server is preferred
+       * when present, otherwise the system server is reached - without the
+       * client having to be told any path. */
+      {
+        const char *runtimeDirectory = getenv("XDG_RUNTIME_DIR");
+
+        if (runtimeDirectory && *runtimeDirectory) {
+          int length = snprintf(xdgPath, sizeof(xdgPath), "%s/brltty", runtimeDirectory);
+          if ((length > 0) && (length < (int)sizeof(xdgPath))) directories[directoryCount++] = xdgPath;
+        }
       }
 
-      if ((sockfd = socket(PF_LOCAL, SOCK_STREAM, 0))<0) {
-        brlapi_errfun="socket";
-        setSocketErrno();
-        goto outlibc;
+      directories[directoryCount++] = BRLAPI_SOCKETPATH;
+
+      if (strcmp(BRLAPI_SOCKETPATH, BRLAPI_SOCKETPATH_LEGACY) != 0) {
+        directories[directoryCount++] = BRLAPI_SOCKETPATH_LEGACY;
       }
+
+      for (directoryIndex=0; directoryIndex<directoryCount; directoryIndex+=1) {
+	const char *directory = directories[directoryIndex];
+	int ldir = strlen(directory);
+
+	if (ldir+1+lport+1 > sizeof(sa.sun_path)) {
+	  brlapi_libcerrno=ENAMETOOLONG;
+	  brlapi_errfun="path";
+	  brlapi_errno = BRLAPI_ERROR_LIBCERR;
+	  continue;
+	}
+
+	if ((sockfd = socket(PF_LOCAL, SOCK_STREAM, 0))<0) {
+	  brlapi_errfun="socket";
+	  setSocketErrno();
+	  goto outlibc;
+	}
 
 #if !defined(HAVE_POLL)
-      if (sockfd >= FD_SETSIZE) {
-	/* Will not be able to call select() on this */
-	closeFileDescriptor(sockfd);
-	brlapi_errfun="socket";
-	setErrno(EMFILE);
-	goto outlibc;
-      }
+	if (sockfd >= FD_SETSIZE) {
+	  /* Will not be able to call select() on this */
+	  closeFileDescriptor(sockfd);
+	  sockfd = -1;
+	  brlapi_errfun="socket";
+	  setErrno(EMFILE);
+	  goto outlibc;
+	}
 #endif /* !HAVE_POLL */
 
-      sa.sun_family = AF_LOCAL;
-      memcpy(sa.sun_path,BRLAPI_SOCKETPATH "/",lpath+1);
-      memcpy(sa.sun_path+lpath+1,port,lport+1);
+	sa.sun_family = AF_LOCAL;
+	memcpy(sa.sun_path,directory,ldir);
+	sa.sun_path[ldir]='/';
+	memcpy(sa.sun_path+ldir+1,port,lport+1);
 
-      if (connect(sockfd, (struct sockaddr *) &sa, sizeof(sa))<0) {
-        brlapi_errfun="connect";
-        setSocketErrno();
-        goto outlibc;
+	if (connect(sockfd, (struct sockaddr *) &sa, sizeof(sa))<0) {
+	  connectError = errno;
+	  closeSocketDescriptor(sockfd);
+	  sockfd = -1;
+	  continue; /* try the next candidate directory */
+	}
+
+	break;
+      }
+
+      if (sockfd < 0) {
+	if (connectError) {
+	  brlapi_errfun = "connect";
+	  errno = connectError;
+	  goto outlibc;
+	}
+
+	/* No directory was actually tried (every candidate was too long); the
+	 * path error that was set still applies. */
+	goto out;
       }
     }
 #endif /* __MINGW32__ */
--- a/Programs/brlapi_client.c
+++ b/Programs/brlapi_client.c
@@ -819,7 +819,6 @@ static int tryHost(brlapi_handle_t *handle, const char *hostAndPort) {
       const char *directories[3];
       unsigned int directoryCount = 0;
       unsigned int directoryIndex;
-      int connectError = 0;
 
       /* Connect to the first server that's found, searching in this order: one
        * the user started within their own session (under $XDG_RUNTIME_DIR), then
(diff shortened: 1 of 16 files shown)


sthibaul:
> BrlAPI is one system daemon

No, that is not true.

We do want people to be able to run their own BrlAPI server. Just like people can run their own X server etc., without having to specify odd paths etc.

alex19EP:
> > BrlAPI is one system daemon
> 
> No, that is not true.
> 
> We do want people to be able to run their own BrlAPI server. Just like people can run their own X server etc., without having to specify odd paths etc.

I don't see this type of deployments in any distribution. but ok. i think we can solve this using xdg run dir witch will be managed by systemd.

alex19EP:
@sthibaul how about this.

- A server started inside a user session publishes its socket under `$XDG_RUNTIME_DIR/brltty/` (i.e. `/run/user/$UID/brltty/`). That directory is already per-user, `0700`, and created/cleaned up by systemd-logind on login/logout — so nothing is world-writable and there's no path to specify, same ergonomics as running your own X server.
- The system server keeps using `/var/lib/BrlAPI`, just at `0755`: it owns the directory and is the only writer, while clients only need to traverse it and connect to the socket (which we already widen to be connectable).
- The client searches `$XDG_RUNTIME_DIR/brltty/<port>` first, then falls back to the system path (and the legacy path for older libbrlapi). So the default local address (empty host / `:0`) still just works — it finds your own session server if you have one, otherwise the system one. No odd paths either way.

Net effect: the world-writable + sticky directory goes away, per-user servers actually get *better* isolation than the shared directory gave them, and existing clients/servers keep working.

The one catch is sequencing: merging the `0755` change on its own would break per-user servers until the XDG discovery lands, so I'd fold the XDG support (server placement + client search order) into this PR rather than merge the hardening alone.

Does that direction sound right? And would you rather keep the system server at `/var/lib/BrlAPI` for backward compat, or move it under `/run` too? I think the latter is more cleaner option.

sthibaul:
> A server started inside a user session publishes its socket under $XDG_RUNTIME_DIR/brltty/

But then if that server is configured to be listening on ':0', it'll create `0` there, and clients will not see the `:0` provided by `root` in `/var/lib/BrlAPI/0` any more, that can confuse users. We have already seen people who start a second instance of brltty (for running an AtSpi driver) and being hit by the BrlAPI driver of brltty connecting to itself and the situation was very confusing and difficult to help debug. If a user starts such a brltty and it creates a server on `:0`, then other clients such as `orca` and such will suddenly connect to that new brltty instead of the root-owned, making again the situation confusing.

The only way I can see to avoid confusion in most situation with this proposal would be to make `--no-api` the default for non-root-started brltty instances. @DaveMielke, what do you think about it? Normal users would usually not need the api when they run brltty themselves. If they do need it, they would probably know they want to use another server number, and it's fine to then ask them to explicitly request starting an api server.

> same ergonomics as running your own X server

AFAICT, user-started X servers are still showing their socket in `/tmp/.X11-unix/` along the others, at least to avoid such confusions.

> would you rather keep the system server at `/var/lib/BrlAPI` for backward compat, or move it under `/run` too?

Moving to `/run` can be useful indeed. `screen` has done that some time ago. This allows to get simple auto-clean at reboot.

> merging the 0755 change on its own would break per-user servers until the XDG discovery lands, so I'd fold the XDG support into this PR rather than merge the hardening alone

sure!

DaveMielke:
[quoted lines by Samuel Thibault on 2026/06/29 at 15:10 -0700]

>The only way I can see to avoid confusion in most situation with this proposal would be to make `--no-api` the default for non-root-started brltty instances. @DaveMielke, what do you think about it? 

I personally think that that's an extremely good idea.

>Normal users would usually not need the api when they run brltty themselves. If they do need it, they would probably know they want to use another server number, and it's fine to then ask them to explicitly request starting an api server.

Yes, exactly!

>> would you rather keep the system server at `/var/lib/BrlAPI` for backward compat, or move it under `/run` too?
>
>Moving to `/run` can be useful indeed. `screen` has done that some time ago. This allows to get simple auto-clean at reboot.

I mhyself like this idea as long as we don't break older clients that don't know about it.

-- 
I recognize the Bible to be the very Word of God: https://Mielke.cc/bible/
Dave Mielke            | 2213 Fox Crescent | Amator et provisor concubitus uxori meae
EMail: ***@***.***  | Ottawa, Ontario   | Home Page: https://Mielke.cc/

The most important document I've ever written: https://mielke.cc/ruler.html
concubitus est ut maritus uxori suae beneplacitum, satisfactionem, et contentum praebeat
cum uxor suaviter tractatur, cunnus eius multo maiorem remissionem mentulae mariti praebebit

alex19EP:
Thanks, @sthibaul and @DaveMielke  agreed on all of it, and `--no-api`-by-default is cleaner than what I proposed. I'll fold this into the PR. Revised plan:

1. Non-root brltty defaults to no-api.
When brltty isn't started by root (`geteuid() != 0`), the embedded server is off unless explicitly turned on (`no-api off` / an explicit server number). A user who genuinely wants their own server enables it and picks a non-zero number, exactly as you said. That kills the `:0` shadowing at the source — no rogue server ever appears at `:0`, so `orca` & co. always reach the root-owned one.

Given that, I'd also **drop the XDG-first client search order** I floated earlier: with no-api as the non-root default there's no per-user `:0` to discover, so an automatic search would just reintroduce exactly the confusion you described. The system server stays the single discoverable one. (XDG for an *explicitly* started user server can be a separate, later thing if anyone ever needs it.) And fair point on `/tmp/.X11-unix/` — I was wrong there. to much Reliance on AI, I guess. Sorry about it.

2. Move the system sockets dir to `/run`.
Default `BRLAPI_SOCKETPATH` → `/run/BrlAPI` (volatile, auto-cleaned at reboot, `0755 brltty:brltty`). To not break older clients still compiled against `/var/lib/BrlAPI` (Dave's constraint), ship a compat symlink `/var/lib/BrlAPI → /run/BrlAPI` (tmpfiles `L` line; the daemon creates it itself on non-systemd). Old clients resolve the same path as before; the new build talks to `/run` directly.

The `0755` hardening already in the PR stays and the above builds on top.

Two things to nail down:
- OK with `/run/BrlAPI` as the name and a `/var/lib/BrlAPI > /run/BrlAPI` symlink as the back-compat shim? Or would you rather the daemon publish into both?
- For the no-api gate, is `geteuid() != 0` the signal you want, or should it key off something else (running under the systemd unit, a config default)? And should the shipped autostart / X-session files set the api mode explicitly so their behavior doesn't shift?

sthibaul:
> I'd also drop the XDG-first client search order I floated earlier: with no-api as the non-root default there's no per-user :0 to discover

We still want non-root users  to be able to easily run a server, that's useful e.g. for easily running an XWindow driver for sighted people to try the braille support of qemu.

alex19EP:
> > I'd also drop the XDG-first client search order I floated earlier: with no-api as the non-root default there's no per-user :0 to discover
> 
> We still want non-root users to be able to easily run a server, that's useful e.g. for easily running an XWindow driver for sighted people to try the braille support of qemu.

ok.

- I'll keep the XDG publish + client search. an explicitly-enabled non-root server publishes to `$XDG_RUNTIME_DIR/brltty/<port>`, and clients search `$XDG_RUNTIME_DIR/brltty/` first, then the system path (`/run/BrlAPI`). So in a session your own clients (orca, etc.) find your server automatically.

Why this no longer hits the confusion you described: that shadowing came from an accidental server grabbing `:0`. With no-api as the non-root default, a session server only exists when you deliberately start one — and then you want your session clients to prefer it. The two `:0`s also live in different directories (`$XDG_RUNTIME_DIR/brltty/0` vs the system `0`), so there's no filesystem collision; it's purely search order. do "prefer my own session server when I've started one" is the behaviour you'd want?

So, just to confirm the precedence:
- when a session (XDG) server exists, session clients prefer it over the system one, else fall back to the system server — good?
- and is a deliberately-started user server at the default `:0` (under XDG) fine, or would you still rather steer such users to a non-zero number?

sthibaul:
> do "prefer my own session server when I've started one" is the behaviour you'd want?

Yes, that will be better.

> when a session (XDG) server exists, session clients prefer it over the system one, else fall back to the system server — good?

Yes

> is a deliberately-started user server at the default :0 (under XDG) fine

Yes.

Programs/brlapi_client.c
```diff
@@ -815,38 +815,86 @@ static int tryHost(brlapi_handle_t *handle, const char *hostAndPort) {
...
+	int ldir = strlen(directory);
+
+	if (ldir+1+lport+1 > sizeof(sa.sun_path)) {
+	  brlapi_libcerrno=ENAMETOOLONG;
+	  brlapi_errfun="path";
+	  brlapi_errno = BRLAPI_ERROR_LIBCERR;
+	  continue;
+	}
+
+	if ((sockfd = socket(PF_LOCAL, SOCK_STREAM, 0))<0) {
```
sthibaul: I believe the socket creation can be kept outside of the loop? If the connect fails, we can reuse the socket to try to connect to the next path

Programs/brlapi_client.c
```diff
@@ -815,38 +815,86 @@ static int tryHost(brlapi_handle_t *handle, const char *hostAndPort) {
...
+	}
+
+	break;
+      }
 
-      if (connect(sockfd, (struct sockaddr *) &sa, sizeof(sa))<0) {
-        brlapi_errfun="connect";
-        setSocketErrno();
-        goto outlibc;
+      if (sockfd < 0) {
```
sthibaul: and you can compare directoryIndex with directoryCount to determine whether we finished the loop early

sthibaul (changes requested):
Sorry for the delay, it seems that github does not notify when PR are updated.

It looks good overall, just a nitpick.

I guess you have tested this with the various cases?

shindere:
Samuel Thibault (2026/07/05 11:17 -0700):
> Sorry for the delay, it seems that github does not notify when PR are
> updated.

Well, there is a setting somewhere to be notified when PR branch gets
pushed but it has the disadvantage that it quickly gets verbose. All in
all I think it's clearer if the PR author comments to let know when an
update occurred.

sthibaul:
thanks, this looks good to me!

DaveMielke:
I'll have to think this one through for a bit. There are two things I'm uncomfortable with.

Introducing an option named "api" while there's already another one named "no-api" is just plain confusing. This hasn't been considered from a user's perspective. They don't even do the same thing. :-) The mere fact that the hack of adding isOptinoWord() should've been a design alarm!

I'm uncomfortable with the changes made to create/ensureDirectory(). Those functions have only been changed with respect to the perceived needs of this specific case. I'm not convinced that the change as it stands won't break something else.

-- 
I recognize the Bible to be the very Word of God: https://Mielke.cc/bible/
Dave Mielke            | 2213 Fox Crescent | Amator et provisor concubitus uxori meae
EMail: ***@***.***  | Ottawa, Ontario   | Home Page: https://Mielke.cc/

The most important document I've ever written: https://mielke.cc/ruler.html
concubitus est ut maritus uxori suae beneplacitum, satisfactionem, et contentum praebeat
cum uxor suaviter tractatur, cunnus eius multo maiorem remissionem mentulae mariti praebebit

DaveMielke:
Merged. Note these changes:

config.c: The --api option hack is gone. The default for --no-api is now set correctly.

cmdline.c: The isWordOption() hack is gone.

brltty@.service: Now sets BRLTTY_NO_API=false rather than BRLTTY_API=true.

The change to create/ensureDirectory() is okay as the sockets directory was the only instance where the second argument was true.

PR #550, opened 2026-07-13
Enhancements to the fr-vs text table

shindere:
- Improve the representation of left and right single quotes
- Define a representation for the capital letter E with grave accent


shindere:
Thanks!

PR #551, opened 2026-07-13
Enhancement to the fr-vs text table

shindere:
This PR makes the representations of the left and right
double quotes coherent with the one of their single counterparts.


shindere:
Thank you again!

PR #561, opened 2026-08-30
bluetooth_darwin: fix classic Bluetooth connectivity, timeouts, and CPU use

### LINKED ISSUE #558: Bluetooth braille displays do not work on macOS
## Summary

On macOS, a braille display connected over classic (BR/EDR) Bluetooth does
not work reliably, and in some cases does not work at all. Tested with a
Freedom Scientific Focus 40 Blue (5th generation); the underlying causes are
in the shared macOS Bluetooth backend (`Programs/bluetooth_darwin.c`), not
specific to this display model.

## Symptoms

- A from-scratch connection attempt can take up to a minute, or fail
  outright, even when the display is powered on, in range, and already
  paired.
- When a connection does succeed, closing it (display powered off, moved out
  of range, or BRLTTY exiting) can leave BRLTTY spinning at 100% CPU
  indefinitely, logging `select error 9: Bad file descriptor` on every
  iteration, instead of shutting down or reconnecting cleanly.
- While any of the above is happening, BRLTTY is also unresponsive in
  general - other braille drivers, the screen-refresh cycle, and anything
  else running in the same process stall for as long as the Bluetooth
  operation is stuck, because of how the macOS backend's blocking waits are
  implemented (see #561 for detail).

## Root causes

Three independent problems in `Programs/bluetooth_darwin.c`, all fixed by
#560 and #561:

1. A stale async input-monitor registration is left pointing at a closed
   file descriptor after a connection closes, which is what causes the
   100%-CPU spin. This is a regression in already-merged #542.
2. RFCOMM channel discovery queries the display before the Bluetooth
   connection is confirmed open, which races a real, still-unfixed
   IOBluetooth bug on modern macOS where the query's completion callback is
   silently dropped.
3. The backend's bounded waits for IOBluetooth callbacks never integrated
   with BRLTTY's own event loop, so a stuck wait froze the entire process,
   and several of those waits were bounded to a full 15-second timeout even
   though a real failure is reported by macOS within a couple of seconds.

## Reference

Part of the broader effort tracked in #540. See #560 and #561 for the
actual fixes and full technical detail.

### LINKED ISSUE #559: Multi-device Bluetooth braille displays connect to the wrong RFCOMM channel on macOS
## Summary

Some braille displays (e.g. the Freedom Scientific Focus line) support
several simultaneous Bluetooth host connections, each its own numbered
slot on the display itself, each potentially assigned a different RFCOMM
channel by the display's own SDP server depending on which slots are
currently in use. On macOS, BRLTTY did not correctly discover this channel
before connecting, so multi-device setups did not work reliably: with only
one host ever paired, the display's default channel and BRLTTY's assumption
happened to line up, masking the problem; with a second host also paired
(e.g. a phone, or macOS's own Accessibility Braille support on the same
Mac), BRLTTY's connection attempt used the wrong channel and failed.

## Symptoms

- A display with only one paired host works over Bluetooth (by coincidence -
  see below).
- Pairing a second host to the same display, or having macOS's own
  Accessibility Braille support connected to it, breaks or destabilizes
  BRLTTY's own Bluetooth connection, even though the display itself shows
  it has a free slot and both connections should be independent.
- The display's own connection-slot indicator does not reliably correspond
  to what BRLTTY is actually able to talk to.

## Root cause

`bthDiscoverChannel()` in `Programs/bluetooth_darwin.c` queried the
display's SDP record for its RFCOMM channel before the underlying Bluetooth
connection was confirmed open. On a cold connection this reliably raced a
still-unfixed IOBluetooth bug (see the companion general Bluetooth issue),
causing discovery to silently fail and BRLTTY to fall back to a hardcoded
channel number - one that only matches the channel this display's *first*
host slot happens to be assigned by default, and stops matching as soon as
a different, or an additional, host is involved.

## Fix

Open (and hold) the Bluetooth connection first, then query for the real
channel - see #561. Verified
live across repeated cold connects: channel discovery now succeeds
reliably and reports the display's actual current channel, not a
hardcoded guess.

## Reference

A similar-looking silent fallback to the wrong channel has also been noticed
with the same kind of display on Linux, but that is a structurally different
code path (`bluetooth_linux.c` already confirms its connection before
querying, so it isn't racing this specific bug) and hasn't been root-caused
or reported separately yet. Part of the broader effort tracked in #540.

hlub:
**Stacked on #560** (the input-monitor fix) - branched on top of it, so this
diff includes that commit until #560 merges, at which point it will shrink
to just the commit below.

## What users get

Classic (BR/EDR) Bluetooth braille displays did not work on macOS at all -
confirmed with a Freedom Scientific Focus 40 Blue, 5th generation. With this
change, the display connects reliably, discovers its actual RFCOMM channel
(needed for the display's own multi-host connection slots to work), and
reconnects quickly after a transient failure (e.g. briefly contending with
macOS's own Accessibility Braille support during a handoff), instead of
hanging for up to a minute. A separate, serious regression is also fixed:
BRLTTY no longer pins a full CPU core indefinitely for the lifetime of a
Bluetooth connection.

## What changed

Several distinct problems, found while getting a real display working end to
end for the first time on this platform:

1. **Channel discovery raced its own connection.** `bthDiscoverChannel()`
   queried the display's SDP record before the Bluetooth connection was
   confirmed open. This races a real, still-unfixed IOBluetooth bug on
   modern macOS where `performSDPQuery:`'s completion callback is silently
   dropped when issued against a not-yet-fully-connected device (confirmed
   live via Console logging Apple's own "This currently won't trigger SDP
   delegate"). Fixed by explicitly opening the connection first
   (`bthEnsureConnectionOpen()`), matching how `bluetooth_linux.c` already
   does this. As a fallback for when the bug still hits an already-open
   link, a synchronous, free cache lookup (`getServiceRecordForUUID:`,
   which only reads records already queried) is tried before falling back
   to a hardcoded channel guess.

2. **Bounded waits didn't integrate with BRLTTY's own event loop.**
   `AsynchronousResult`'s `wait:` (used throughout this file for every
   async IOBluetooth call) only pumped this thread's CoreFoundation run
   loop directly. It never touched BRLTTY's own cross-platform event loop
   (`async_wait.c`'s `asyncAwaitCondition()`/`awaitAction()`, which
   services alarms and other file descriptors). Confirmed live: while
   blocked in any of these waits, everything else in the single-threaded
   process - other drivers, the screen-refresh alarm, a companion helper
   process's readiness signal - was frozen for as long as the wait ran. On
   every other platform, the equivalent blocking wait inside a driver
   callback (e.g. `bluetooth_linux.c`'s connect wait) is a *recursive*
   re-entry into that same event loop, so alarms and I/O keep running while
   it blocks. `wait:` now does the same, looping `asyncAwaitCondition()` in
   small slices so the CFRunLoop is drained regularly regardless of what
   else happens to be scheduled.

3. **Failed steps waited out the full connect timeout for no reason.** A
   real failure at the connection-open, SDP-query, or RFCOMM-open step is
   reported by macOS's own Bluetooth stack within a couple of seconds at
   most, but the completion callback for that failure is not reliably
   delivered (same underlying bug as #1). Each of these three steps now
   uses a much shorter, Darwin-specific timeout for its own wait, rather
   than the full 15-second cross-platform connect timeout every other
   platform and step still needs for genuinely slow first-time connects.

4. **A CPU-use regression from fix #2's own supporting infrastructure.** A
   periodic timer that drains the CoreFoundation run loop (needed so
   IOBluetooth can deliver incoming data at all - see the
   `runLoopPumpAlarm` field comment) ran every 20ms for the entire lifetime
   of a connection. That interval was originally chosen to speed up
   connection *establishment*, before fix #2 existed; once fix #2 landed,
   nothing needs it that tight any more, but it kept running just as often
   during ordinary, mostly-idle interactive use. Confirmed live: pinned a
   full CPU core at 100%, including long idle stretches with zero
   Bluetooth activity. Slowed to 100ms, still responsive for interactive
   key presses.

5. **A dead background thread, and the real race it was hiding.** Once an
   RFCOMM channel opened, a dedicated background thread was spawned whose
   only job was to keep its own run loop pumped, on the theory that
   IOBluetooth might deliver incoming-data callbacks there. Confirmed live
   with temporary thread-identity logging across multiple full connections:
   every single callback arrived on the main thread instead (the same
   `runLoopPumpAlarm` mechanism from fix #4 already pumps that) - the
   background thread did nothing but consume a thread's worth of resources
   for the life of every connection. Removing it also closed a real,
   previously-unnoticed race: the handoff of the channel's delegate to this
   thread happened *asynchronously*, whenever it next got scheduled, so a
   callback arriving in that gap could still have gone to the previous
   (open-completion) delegate. The handoff is now synchronous, immediately
   after the channel opens, closing that gap. Two smaller, live-tested
   simplifications landed alongside this: an unauthenticated connection
   request now behaves identically to an authenticated one for security
   purposes (macOS enforces its own encryption policy for this service
   class regardless - confirmed by comparing Console logs side by side for
   both), and a leftover fake serial-port-configuration command sent after
   opening the channel was confirmed unnecessary (the display works
   identically without it) and removed.

## Platform scope

macOS / IOBluetooth backend only (`Programs/bluetooth_darwin.c`,
`Programs/system_darwin.c`, `Headers/system_darwin.h`). The one small,
Bluetooth-specific helper that must live in `Programs/bluetooth.c` instead
(a plain-C file, needed because of a `TimeValue` symbol collision between
this codebase's own headers and the Foundation/CoreFoundation headers the
Objective-C files require) is documented in place. No other platform's
Bluetooth backend is touched.

**Not included here, deliberately:** a short `Documents/README.macOS`
section on pairing a Bluetooth braille display through VoiceOver Utility
rather than System Settings (and the two-step removal order needed to
fully clear a stale pairing when re-pairing) was written and tested
alongside this fix, but `README.macOS` doesn't exist on `master` yet - it
ships with the separate, not-yet-opened `brltty-handoff`/macOS-docs PR. The
doc change lives on this fork's own `macos/integration` branch for now and
should be folded into that PR whenever it's opened, not into this one.

## Testing / risk

Extensively tested live against real hardware (a Focus 40 Blue) across many
connect/disconnect cycles, including deliberately stalled attempts (an
unreachable address) to confirm other BRLTTY activities keep running during
a Bluetooth wait, and repeated real-world handoffs with macOS's own
Accessibility Braille support contending for the same device. This is the
first time this backend's actual data-exchange path (not just the
bounded-failure path) has been exercised on real hardware. Verified both as
a standalone foreground process (used for most of the above, for fast
iteration) and through the full real daily-use path (`sudo` +
`brltty-term`/`brltty-handoff` + `brltty-pty`).

## Reference

Supersedes two earlier, never-opened local drafts on this fork
(`bluetooth_darwin: honor connect and name-lookup timeouts` and `bluetooth:
don't wipe the connect-error cache on every driver-activation retry`), both
held pending hardware access. Testing this change in full superseded both:
the timeout-conversion idea from the first is carried forward in a more
complete form (the event-loop integration in #2 above), and the retry-cache
idea from the second turned out, once real attempts became cheap via #3, to
cost more than it saved (confirmed live: a since-cleared transient failure
still had to wait out a stale 30-second cache window before the next real
attempt). It was tried, retuned, and ultimately removed rather than shipped
in any form - the `config.c`/`bluetooth.c` state this PR leaves behind is
therefore identical to current `master`, not a new change; no PR is being
opened for it.

Closes #559 (multi-device wrong-channel issue) on its own. Together with
#560, fully fixes #558 (general macOS Bluetooth issue) - not marked as
"Closes" here since #558 needs both this PR and #560 merged, not this one
alone.

Part of the umbrella tracking issue for first-class braille access in GUI
terminals and macOS support (#540).

### DIFF
--- a/Drivers/Braille/FreedomScientific/braille.c
+++ b/Drivers/Braille/FreedomScientific/braille.c
@@ -848,7 +848,7 @@ brl_construct (BrailleDisplay *brl, char **parameters, const char *device) {
     if (connectResource(brl, device)) {
       FS_Packet response;
 
-      if (probeBrailleDisplay(brl, 2, NULL, 100,
+      if (probeBrailleDisplay(brl, 2, NULL, 1000,
                               writeIdentifyRequest,
                               readResponse, &response, sizeof(response),
                               isIdentityResponse)) {
--- a/Headers/system_darwin.h
+++ b/Headers/system_darwin.h
@@ -21,13 +21,14 @@
 
 #include <CoreFoundation/CFRunLoop.h>
 
-#import <Foundation/NSThread.h>
+#include "async_types_handle.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif /* __cplusplus */
 
 extern IOReturn executeRunLoop (int seconds);
+extern void darwinDrainRunLoop (void);
 extern void addRunLoopSource (CFRunLoopSourceRef source);
 extern void removeRunLoopSource (CFRunLoopSourceRef source);
 
@@ -38,22 +39,12 @@ extern void setDarwinSystemError (IOReturn result);
 @property (assign, readonly) int isFinished;
 @property (assign, readonly) IOReturn finalStatus;
 
+/* timeoutMilliseconds: how long to wait for isFinished to become true. */
 - (int) wait
-  : (int) timeout;
+  : (int) timeoutMilliseconds;
 
 - (void) setStatus
   : (IOReturn) status;
-@end
-
-@interface AsynchronousTask: AsynchronousResult
-@property (assign, readonly) NSThread *taskThread;
-@property (assign, readonly) CFRunLoopRef taskRunLoop;
-
-- (IOReturn) run;
-
-- (int) start;
-
-- (void) stop;
 @end
 
 #ifdef __cplusplus
--- a/Programs/bluetooth.c
+++ b/Programs/bluetooth.c
@@ -32,6 +32,11 @@
 #include "io_bluetooth.h"
 #include "bluetooth_internal.h"
 
+#ifdef __APPLE__
+#include "async_alarm.h"
+#include "async_handle.h"
+#endif /* __APPLE__ */
+
 static int
 bthDiscoverSerialPortChannel (uint8_t *channel, BluetoothConnectionExtension *bcx, int timeout) {
   static const uint8_t uuid[] = {
@@ -634,6 +639,15 @@ bthGetDeviceName (uint64_t address, int timeout) {
 
   if (entry) {
     if (!entry->name) {
+      /* Deliberately independent of bthOpenConnection()'s connect-error
+       * cache: a name request and a channel connect are different
+       * operations against the same address, and can fail for different
+       * reasons (e.g. a page-scan timing quirk can fail a name request
+       * while the actual SPP channel would still connect fine). Sharing
+       * one cache between them let a failing name request silently starve
+       * every subsequent real channel-connect attempt for the address, for
+       * as long as the connect-error cache stayed unforgotten - the real
+       * connection was then never even attempted. */
       logMessage(LOG_CATEGORY(BLUETOOTH_IO), "obtaining device name");
 
       if ((entry->name = bthObtainDeviceName(address, timeout))) {
@@ -785,3 +799,37 @@ int
 isBluetoothDeviceIdentifier (const char **identifier) {
   return hasQualifier(identifier, BLUETOOTH_DEVICE_QUALIFIER);
 }
+
+#ifdef __APPLE__
+/* macOS's IOBluetooth delivers its asynchronous callbacks (including
+ * incoming RFCOMM data) by scheduling a source on whatever thread's run
+ * loop was current when the connection was made - here, this process's
+ * single main thread - and nothing else in BRLTTY ever pumps that run
+ * loop, so those callbacks are otherwise never delivered. This lives here
+ * rather than in bluetooth_darwin.c because that file is an Objective-C
+ * translation unit pulling in Foundation, whose MacTypes.h typedefs a
+ * TimeValue that collides with this codebase's own (Headers/timing_types.h,
+ * pulled in by async_alarm.h). Not declared via system_darwin.h either,
+ * for the same reason: that header pulls in the same conflicting
+ * Foundation/CoreFoundation headers. */
+extern void darwinDrainRunLoop (void);
+
+ASYNC_ALARM_CALLBACK(bthDarwinRunLoopPumpAlarmCallback) {
+  darwinDrainRunLoop();
+}
+
+void
+bthStartDarwinRunLoopPump (AsyncHandle *handle) {
+  if (asyncNewRelativeAlarm(handle, DARWIN_BLUETOOTH_RUN_LOOP_PUMP_INTERVAL, bthDarwinRunLoopPumpAlarmCallback, NULL)) {
+    if (asyncResetAlarmInterval(*handle, DARWIN_BLUETOOTH_RUN_LOOP_PUMP_INTERVAL)) return;
+    asyncCancelRequest(*handle);
+    *handle = NULL;
+  }
+
+  /* Without this alarm, incoming Bluetooth data is silently never
+   * delivered at all (see the comment above) - if it couldn't even be
+   * created or repeated, that failure needs to be visible rather than
+   * quietly degrading back into that exact bug. */
+  logMessage(LOG_WARNING, "could not start Bluetooth run loop pump alarm - incoming data may never be received");
+}
+#endif /* __APPLE__ */
--- a/Programs/bluetooth_darwin.c
+++ b/Programs/bluetooth_darwin.c
@@ -27,6 +27,7 @@
 #import <IOBluetooth/objc/IOBluetoothRFCOMMChannel.h>
 
 #include "log.h"
+#include "parameters.h"
 #include "io_misc.h"
 #include "io_bluetooth.h"
 #include "bluetooth_internal.h"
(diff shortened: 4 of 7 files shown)


hlub:
Added one more commit: split `DARWIN_BLUETOOTH_ASYNC_STEP_TIMEOUT` into three independent per-step constants.

Live testing after this PR's own fix showed the RFCOMM-open step specifically hits the dropped-callback failure on essentially every reconnect right after another local Bluetooth client releases the display - not an occasional flake - so its 3s timeout was routinely paid as pure dead time. The other two async steps (connection-open, SDP query) haven't shown that same near-certain pattern, so rather than lowering the shared constant on one step's evidence, it's now three independent constants: RFCOMM-open tuned aggressively (1500ms), the other two left at the original conservative 3000ms.

DaveMielke:
I have a problem with a PR such as this one because, as its log even says, it's tackling several problems at the same time. That keeps me from dealing with each one individually.

I have no problem with the fs driver change, i.e. just increasing a failure timeout, so I think I'll just go ahead and make that change myself.

I do have a problem with the main change, though, as I don't particulrly like platform-specific code being sprinkled throughout portable code. We've worked very hard to minimize the need for that.

If I'm understanding it correctly, the problem is that certain Apple events can't be handled by brltty's async capability. It looks like Apple's run loop needs to be started, and then to be continaully polled for events. I'm guessing that, once polled, all pending events are run and then the poll call returns.

Why not start the run loop in system_darwin.c:initializeSystemObject(), and then have Darwin-specific code repoll as needed?

hlub:
Thanks for the review - you're right on both counts, and I've reworked this.

**Split up.** The fs driver change is gone (you already made it yourself, separately). The SDP-query-ordering fix (display's SDP record queried before its connection was confirmed open) is now its own PR, #569 - it doesn't depend on anything here, it's built directly on current master. What's left in this PR is one story: the async wait/timeout/CPU-pump mechanism.

**On the platform code in portable files:** you're right that it shouldn't be there, and I've moved it out - bluetooth.c and bluetooth_internal.h are now identical to master again, zero platform-specific code.

I looked into your suggestion of putting the run-loop pump in system_darwin.c:initializeSystemObject() specifically, and it can't quite go there as stated. The pump needs to schedule a repeating alarm (async_alarm.h), but async_alarm.h pulls in this codebase's own struct TimeValue (Headers/timing_types.h), and that collides with a typedef of the same name in Apple's own CoreFoundation headers - any file that includes CoreFoundation can't also include async_alarm.h, they fail to compile together. system_darwin.c already includes CoreFoundation (for CFRunLoopRef), so it hits the same wall bluetooth_darwin.c does; that's the actual reason this ended up somewhere unexpected in the first place.

What I did instead: a new, small, plain-C file (system_darwin_runloop.c) that never touches CoreFoundation at all, wired into the build the same way HOSTCMD_OBJECTS already rides along with SYSTEM_OBJECTS today - one main per-platform object plus a companion object, no new build concept. It exposes darwinRequestRunLoopPump(), declared in system_darwin.h alongside the rest of the Darwin system layer, so bluetooth_darwin.c just calls into "the Darwin system layer" the same way it already does for everything else there - not something Bluetooth invented for itself. It's still scoped to the lifetime of an actual Bluetooth connection rather than running for the whole life of the process, so a process that never touches Bluetooth (which is most of the time even for me, since I run USB as primary) pays nothing for it.

Also dropped the now-redundant input-monitor commit from this branch's history - that one's already merged as #560, and rebasing onto current master picked that up cleanly.

Re-verified with a clean ./autogen + ./cfg-darwin + make build on macOS after all of the above, and re-tested live across repeated connect/disconnect cycles.

hlub:
Ran an independent review pass against this and #569 together, since they touch adjacent code. It caught a real problem worth flagging here: my "split" wasn't actually clean - this PR still had its own copy of the connection-open-before-SDP-query fix that #569 also implements, which would have collided once both are looked at together. Removed it entirely from this PR; that fix now lives only in #569.

It also caught two real inconsistencies in this PR's own stated design: ServiceQueryResult was released even on a timed-out wait, unlike every other async result type added in this same commit (which deliberately leak on timeout rather than risk a late callback touching freed memory) - fixed to match. Same issue for RfcommChannelDelegate at connection teardown, released instead of abandoned like RfcommChannelOpenObserver already is for the identical situation - fixed to match, with a nil-guard added to its data callback.

Also stripped some leftover PTY-related constants and an unrelated update-delay change that had no business being in this file - not sure how those survived this far, but they're gone now.

Rebuilt clean (./autogen + ./cfg-darwin + make) after all of the above.

PR #563, opened 2026-09-01
brlapi_server: retry the full device list when a resume attempt fails

hlub:
## Summary

When a BrlAPI client resumes a suspended braille driver, a failed resume attempt could leave the driver stuck permanently unconstructed, with nothing ever retrying - and a stale per-address connect-error cache could silently keep it that way even after the underlying device becomes reachable again.

## What changed

Two related gaps in the resume path (`Programs/brlapi_server.c`):

1. `resumeBrailleDriver()` reconnects only to the single device that was active before suspend, via `constructBrailleDriver()` directly - unlike a normal cold start (`startBrailleDriver()`, `config.c`), which walks the full configured device list. If that one device is unreachable when resume happens, there's no fallback to any other configured device.

2. Both callers of this resume path - `handleResumeDriver()` (the normal `RESUMEDRIVER` packet handler) and the safety net for a client that disconnects without resuming cleanly - previously discarded a failed resume entirely. Confirmed live: this leaves the driver stuck unconstructed indefinitely after a single failed attempt, since resume isn't on any retry timer the way a cold start is.

Compounding both: `resumeBrailleDriver()` never called `forgetDevices()` before attempting, so a per-address connect-error cache entry left over from before the suspend (or from an earlier failed resume on this same path) can silently short-circuit the very next real attempt too, without even trying. Normal cold starts always clear this cache first; this path never did.

Fixed by having `resumeBrailleDriver()` forget devices before attempting (so a resume is always a genuine attempt), and by falling back to a full `restartBrailleDriver()` - which retries the complete configured device list - whenever the narrow resume attempt fails, on both call sites.

## Platform scope

Cross-platform (`Programs/brlapi_server.c`, `Programs/config.c` untouched, no header changes). Not specific to any one driver or backend - applies to any setup using BrlAPI's suspend/resume mechanism (e.g. handing a display over to another consumer temporarily and back).

## Testing / risk

Found and reproduced on macOS, in a setup that suspends/resumes BRLTTY's Bluetooth connection to hand a display to the OS's own accessibility braille support and back. Confirmed live: after this fix, a failed resume attempt now recovers automatically within a few seconds via the fallback, instead of leaving the driver stuck indefinitely - watched this happen multiple times across a testing session. The fallback path (`restartDriver()`) reuses the exact core-task/mutex pattern the existing resume path already used, so no new locking discipline is introduced; verified it doesn't deadlock by exercising it live repeatedly.

## Reference

Found while testing macOS Bluetooth reconnection reliability - part of the broader effort tracked in #540.

### DIFF
--- a/Programs/brlapi_server.c
+++ b/Programs/brlapi_server.c
@@ -409,6 +409,19 @@ static int resumeBrailleDriver(BrailleDisplay *brl) {
   if (trueBraille == &noBraille) return 0; /* core unlinked api */
   driverConstructing = 1;
   lockMutex(&apiSuspendMutex);
+
+  /* A resume attempt must be a genuine attempt, not one silently answered
+   * from a stale per-address connect-error cache entry left over from
+   * before the suspend (bthCloseConnection() never clears it) or from an
+   * earlier failed resume (nothing on this path ever calls it either).
+   * Without this, a single connect failure here - e.g. the still-unfixed
+   * IOBluetooth RFCOMM race documented in bluetooth_darwin.c - permanently
+   * short-circuits every later resume for this device: bthOpenConnection()
+   * would keep honoring the cached failure and never even try again. The
+   * normal cold-start path (startBrailleDriver(), config.c) already does
+   * this before every attempt; this path never did. */
+  forgetDevices();
+
   driverConstructed = constructBrailleDriver();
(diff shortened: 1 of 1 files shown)


sthibaul:
This looks good to me, thanks!

PR #569, opened 2026-09-18
bluetooth_darwin: open the connection before querying its SDP record

hlub:
Split out of #561 as its own, independently-reviewable fix: the display's SDP record was being queried before its Bluetooth connection was confirmed open, which races a real IOBluetooth bug where the SDP-query completion callback silently never arrives on a not-yet-connected device.

This one does not depend on #561 at all - it's built directly on top of current master, using the same async-waiting mechanism that's already there. See the commit message for the full root-cause and fix description.

Confirmed with a Freedom Scientific Focus 40 Blue, 5th generation: this bug specifically broke the display's multi-host connection slots, since a failed discovery silently fell back to a hardcoded channel guess that only happened to be right when this Mac was the display's only connected host.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TA2HUPZNUFSL9Zw7hBF8oz


hlub:
Ran an independent review pass on this. Two real findings, both fixed:

- The cache-lookup fallback (getServiceRecordForUUID:) ran unconditionally, even after the live SDP query got a definitive negative answer rather than just a dropped callback. That risked returning a stale cached channel instead of correctly falling through to the caller's own hardcoded fallback. Now it's skipped on a real failure, only tried on success or on a timeout/dropped callback.
- A comment claimed calling openConnection on an already-connected device is idempotent success - the actual IOBluetooth header says the opposite since macOS 10.7 (that masking was explicitly removed). Fixed the comment, and added an honest note that this call has no BRLTTY-imposed timeout of its own, only IOBluetooth's default page timeout - a deliberate, accepted one-time-per-attempt cost, not something silently assumed to be instant.

Also caught and fixed the same release-on-timeout inconsistency in ServiceQueryResult that turned up in #561's review (this file's copy of that class had the identical issue).

Rebuilt clean (./autogen + ./cfg-darwin + make) after the fixes.

